PR #95, opened 2021-11-09
Post build plugin

ivarpruijn:
Copies .htaccess to build directory after building, for both website and docs.


docs/docs/tracking/introduction.md
```diff
@@ -28,13 +28,13 @@ import Mermaid from '@theme/Mermaid';
...
-    click Event href "core-concepts/events" "See more details" _self
-    click Location href "core-concepts/locations" "See more details" _self
-    click Collector href "core-concepts/collector" "See more details" _self
+    click Tracker href "/docs/tracking/core-concepts/trackers" "See more details" _self
+    click Tagging href "/docs/tracking/core-concepts/tagging" "See more details" _self
+    click Plugins href "/docs/tracking/core-concepts/trackers#extensibility-plugins" "See more details" _self
+    click Validation href "/docs/tracking/core-concepts/validation" "See more details" _self
+    click Event href "/docs/tracking/core-concepts/events" "See more details" _self
+    click Location href "/docs/tracking/core-concepts/locations" "See more details" _self
+    click Collector href "/docs/tracking/core-concepts/collector" "See more details" _self
```
sdirosa, line 37: I'm curious:
Is `/docs` needed here? The baseUrl is `/docs` already, right? 

Shouldn't  this should work fine with just `/tracking`?

docs/docusaurus.config.js
```diff
@@ -14,14 +14,14 @@ const config = {
   titleDelimiter: '|',
   tagline: 'Objectiv is a data collection & modeling library that puts the data scientist first.',
   url: 'https://objectiv.io/',
-  baseUrl: '/',
+  baseUrl: (environment==='prod') ? '/docs/' : '/',
```
sdirosa, line 17: I'd use `environment.startsWith('prod)` so if someone sets `production` by mistake, it still catches it. 


Extra, maybe nicer:

```javascript
const isProductionEnv = () => {
  const environment = process.env.OBJECTIV_ENVIRONMENT;
  return environment ? environment.startsWith('prod') : false;
}
```

docs/src/plugins/post-build/index.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs');
+
+module.exports = function (context, options) {
+  return {
+    name: 'add-htaccess',
+    async postBuild({siteConfig = {}, routesPaths = [], outDir}) {
+      // copy .htaccess from root to ./build
+      // destination will be created or overwritten by default.
+      fs.copyFile('./.htaccess', './build/.htaccess', (err) => {
```
sdirosa, line 9: To make this a little more solid (across platforms), use the `path` library. It will ensure paths are built with the correct slashes and dot notations for the target OS.

```javascript
const path = require('path');
```

```javascript
fs.copyFile('.htaccess', path.join('build', '.htaccess')'', (err) => {
  ...
}

```

src/css/custom.css
```diff
@@ -150,6 +150,13 @@ div[role=banner] img {
   }
 }
 
+@media (min-width: 1440px) {
+  /* fix: Docusaurus builds set container max-width to var(--ifm-container-width-xl) on large screens */
+  .container {
+    max-width: 996px !important;
+  }
+}
+
```
sdirosa, line 159: Nice find!

src/plugins/post-build/index.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs');
+
+module.exports = function (context, options) {
+  return {
+    name: 'add-htaccess',
+    async postBuild({siteConfig = {}, routesPaths = [], outDir}) {
+      // copy .htaccess from root to ./build
+      // destination will be created or overwritten by default.
+      fs.copyFile('./.htaccess', './build/.htaccess', (err) => {
```
sdirosa, line 9: Same here

sdirosa (changes requested):
I left a couple of minor improvement suggestions 👌🏼

docs/docs/tracking/introduction.md
ivarpruijn: I don't think docusaurus appends that baseUrl here within the Mermaid syntax. So if we would use `/tracking` instead of `docs/tracking`, it would lose the `/docs`.

One way to fix this is to extend our Mermaid component to provide it a list of clickable elements (with their URL, tooltip and target), and let it generate this syntax including the `baseUrl`. Would be more future-proof, if we ever move the docs to another URL.

sdirosa: That sounds like a good improvement! 

I was wondering how would we keep track of these URLs exactly in a scenario as you described. I guess we could still grep /tracking and find them, but it's not ideal.

Thanks for explaining how this works

docs/docusaurus.config.js
ivarpruijn: Nice one, just committed the change

docs/src/plugins/post-build/index.js
ivarpruijn: Done & committed

docs/docs/tracking/introduction.md
ivarpruijn: I'll put the improvement on my list!

ivarpruijn:
@borft Merging it now after review from @sdirosa , please let me know if you find anything afterwards, e.g. in the `.htaccess` files.

PR #103, opened 2021-11-10
Docs - modeling: fix anchor scroll offset

ivarpruijn:
Simply setting the navbar height to an absolute value should do the trick.


sdirosa (approved):
I think this is already an improvment 👌🏼

borft:
lgtm. Still weird this behaves differently on dev vs prod

PR #105, opened 2021-11-11
Docs section in mobile menu

ivarpruijn:
Currently, when you land on any of the docs on a mobile device, there's no indication that there are multiple Sections (Home, Tracking, Modeling, Taxonomy). This is because the top navigation bar is not directly shown anywhere, including the hamburger menu; you have to first choose 'Back to main menu':

image

We want to make sure that, no matter the entry point, users can understand which section they are, and that there are others.

This PR changes the UX in two ways:
- 'Back to main menu' in the hamburger menu is now '← All Sections'.
- The hamburger menu now shows the section of the Docs you're in:
image


docs/src/theme/Navbar/index.tsx
```diff
@@ -148,6 +149,19 @@ function NavbarMobileSidebar({sidebarShown, toggleSidebar}) {
...
     toggleSidebar,
   });
+
+  // Determine active Docs section
+  const {leftItems} = splitNavItemsByPosition(useNavbarItems());
+  let activeDocsSection = '';
+  for (const item of leftItems) {
+    const toUrl = useBaseUrl(item.to);
+    // strip last '/' from end of URL
+    const toUrlStripped = toUrl.substring(0, toUrl.length-1);
```
ivarpruijn, line 159: This is basically a workaround for the double slashes we currently use for the URLs in the top nav

sdirosa (approved):

PR #107, opened 2021-11-11
WIP: Render only main section from generated bach pages

sdirosa:
@ivarpruijn This works, but it doesn't solve the offset issue either.

### DIFF
--- a/docs/src/components/sphinx-page/index.tsx
+++ b/docs/src/components/sphinx-page/index.tsx
@@ -194,7 +194,10 @@ const SphinxPage = (props) => {
                     }
                 }
 
-                const data = tempDiv.innerHTML;
+                // Extract the main section only, we don't need the wrapping <div>s and <main> tags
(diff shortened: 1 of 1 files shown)


docs/src/components/sphinx-page/index.tsx
```diff
@@ -194,7 +194,10 @@ const SphinxPage = (props) => {
                     }
                 }
 
-                const data = tempDiv.innerHTML;
+                // Extract the main section only, we don't need the wrapping <div>s and <main> tags
+                const mainSection = tempDiv.querySelector('section');
```
sdirosa, line 198: The main tag is currently not closed in the generated pages. At least this gets rid of that

ivarpruijn:
@sdirosa I think we can close this PR now, right?

sdirosa:
> @sdirosa I think we can close this PR now, right?

I think so, haven't double checked, but I think Mike mentioned fixing this while refactoring the docs

PR #109, opened 2021-11-11
Docs: Restructure Taxonomy docs

ivarpruijn:
This PR is a proposal to restructure the Taxonomy section to be inline with the Tracking section:
- An 'Introduction' instead of 'Overview'.
- A Core Concepts category.
- All Taxonomy items moved to a 'Reference' category.
- All links to the Taxonomy items updated to use the new structure.


jansenbob:
LGTM. 

Two small notes:

- Maybe add quick link(s) to core concepts from introduction as well
- A single file for core concepts might be sufficient

sdirosa (approved):
Looks like a nice restructuring to me, I like how it aligns a bit with other sections

ivarpruijn:
> LGTM.
> 
> Two small notes:
> 
> * Maybe add quick link(s) to core concepts from introduction as well
Definitely, content there is WIP.

> * A single file for core concepts might be sufficient
Think you're right that it would be short enough, so makes sense. Only downside is that it's not 100% consistent with the Tracking section, and possibly later the Modeling section, but not sure that's much of an issue.

jansenbob:
> > LGTM.
> > Two small notes:
> > 
> > * Maybe add quick link(s) to core concepts from introduction as well
> 
> Definitely, content there is WIP.
> 
> > * A single file for core concepts might be sufficient
> 
> Think you're right that it would be short enough, so makes sense. Only downside is that it's not 100% consistent with the Tracking section, and possibly later the Modeling section, but not sure that's much of an issue.

I don't think that's an issue

ivarpruijn:
@jansenbob Discussed with @vincenthoogsteder : structure looks good, and we'll turn the Core Concepts into a single file as you mentioned. I'll write a proposal for the content and will ping you for a review.

jansenbob (approved):

vincenthoogsteder (approved):

PR #112, opened 2021-11-12
Docs: Mermaid safe links

ivarpruijn:
Pass link clicks in Mermaid diagrams on to the component, and let it handle URLs with docusaurus' `useBaseUrl`. In the future we could also track the link clicks.

I've only updated the diagrams in the Home and Tracking sections. Will update the ones used in the Taxonomy section in PR #109.

### DIFF
--- a/docs/src/theme/Mermaid/index.tsx
+++ b/docs/src/theme/Mermaid/index.tsx
@@ -46,22 +46,20 @@ const Mermaid = ({ chart, caption, baseColor, links}) => {
 		mermaid.contentLoaded();
 	}, []);
 
-  let clicks = '';
   if(links && links.length > 0) {
     for (const link of links) {
-      // todo: use baseUrl
       const name = link.name ? link.name : '';
+      // use the base URL for all links
       const linkTo = link.to ? useBaseUrl(link.to) : '#';
       const tooltip = link.tooltip ? link.tooltip : 'See details';
       const target = link.target ? link.target : '_self';
-      clicks = clicks + '\n click '+name+' "'+linkTo+'" "'+tooltip+'" '+target+'';
+      chart = chart + '\n click ' + name + ' "' + linkTo + '" "' + tooltip + '" ' + target + '';
     }
   }
-  const chartClicks = chart + clicks;
 
 	return (
     <div className={"diagram-"+baseColor}>
-      <div className="mermaid">{chartClicks}</div>
+      <div className="mermaid">{chart}</div>
       <p className="diagram-caption">{caption}</p>
     </div>
   );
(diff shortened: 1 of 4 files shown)


docs/src/theme/Mermaid/index.tsx
```diff
@@ -40,11 +41,28 @@ mermaid.initialize({
...
+      // use the base URL for all links
+      const linkTo = link.to ? useBaseUrl(link.to) : '#';
+      const tooltip = link.tooltip ? link.tooltip : 'See details';
+      const target = link.target ? link.target : '_self';
+      chart = chart + '\n click ' + name + ' "' + linkTo + '" "' + tooltip + '" ' + target + '';
+    }
+  }
+
+	return (
+    <div className={"diagram-"+baseColor}>
```
ivarpruijn, line 61: Here we could also add tracking in the future.

sdirosa: Curious: have you already tried tagging the div and see if it works?

sdirosa (approved):
I've left a couple of questions only, good to go for me

docs/src/theme/Mermaid/index.tsx
```diff
@@ -40,11 +41,28 @@ mermaid.initialize({
...
 	}, []);
-	return <div className={"diagram-"+baseColor}><div className="mermaid">{chart}</div><p className="diagram-caption">{caption}</p></div>;
+
+  if(links && links.length > 0) {
+    for (const link of links) {
+      const name = link.name ? link.name : '';
+      // use the base URL for all links
+      const linkTo = link.to ? useBaseUrl(link.to) : '#';
+      const tooltip = link.tooltip ? link.tooltip : 'See details';
+      const target = link.target ? link.target : '_self';
```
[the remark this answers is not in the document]
ivarpruijn: The nullish-coalesce syntax looks good, I'll update it.

And agree that the `links.to` should not be optional now I think about it. I'll make a change to skip the link when it's not present.

docs/src/theme/Mermaid/index.tsx
ivarpruijn: Not yet. But no reason why it would not work I think, I'll give it a try.

ivarpruijn: Tagging the div works: 
image

So I think we could also tag the Links in it using `tagChildren`, but I'll leave that for another time.

docs/src/theme/Mermaid/index.tsx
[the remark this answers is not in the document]
ivarpruijn: Updated, please have a look @sdirosa

docs/src/theme/Mermaid/index.tsx
sdirosa: Pretty nice, that makes quite easy actually

docs/src/theme/Mermaid/index.tsx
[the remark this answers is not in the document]
sdirosa: looks good!

sdirosa:
I'd say: merge it!

PR #113, opened 2021-11-12
Enable trailingSlash setting

ivarpruijn:
Enables the `trailingSlash` setting, so that links like https://objectiv.io/privacy/ will not result in a directory listing, but show the intended page instead.

Need to test carefully whether this breaks any links to the standalone docs.


ivarpruijn:
Superseded by #115

PR #116, opened 2021-11-12
Ensure docusaurus redirects to the standalone doc

ivarpruijn:
Last minute change I forgot to commit.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -84,7 +84,7 @@ const config = {
       items: [
         {
           label: 'Docs',
-          to: '/docs/',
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -84,7 +84,7 @@ const config = {
       items: [
         {
           label: 'Docs',
-          to: '/docs/',
+          to: 'https://objectiv.io/docs/', // ensure Docusaurus redirects to standalone docs
           target: '_self'
```
sdirosa, line 88: Ah, nice one

PR #120, opened 2021-11-15
Use NODE_ENV to determine configuration values

ivarpruijn:
Provides a failsafe for creating builds:
- On `yarn start`, always uses development config for all tracker values such as the Collector address.
- On `yarn build`, always uses production config for all tracker values such as the Collector address and Docusaurus config.
- When `OBJECTIV_ENVIRONMENT` environment variable is set to `staging`, the corresponding values are used for the tracker and Docusaurus config.


docs/docusaurus.config.js
```diff
@@ -100,7 +100,7 @@ const config = {
           {
             label: 'Objectiv.io \u{1F855}',
             position: 'right',
-            to: 'https://objectiv.io/',
+            to: isStagingEnv ? 'https://staging.objectiv.io/' : 'https://objectiv.io/',
```
sdirosa, line 103: Question: 

Why do we have to use absolute URLs here?

ivarpruijn: If we use relative URLs, Docusaurus/React treats them as such, and tries to load `/docs` instead.

sdirosa: Isn't that because of the Link components?

sdirosa: Just wondering if we can solve this in another cleaner way btw

ivarpruijn: Think so, yes. But if we would change the Navbar component to only use <a> links instead (yet another change to that component), it would open any link in the navbar by default in a new tab. So then we'd have to set `target: '_self'` to every navbar link in the configuration. Bit of a tradeoff, as we now run the docs standalone.

docusaurus.config.js
```diff
@@ -3,9 +3,10 @@
 
 const path = require('path');
 
-const environment = process.env.OBJECTIV_ENVIRONMENT;
-const isStagingEnv = environment ? environment.startsWith('staging') : false;
-const isProductionEnv = environment ? environment.startsWith('prod') : false;
+const nodeEnv = process.env.NODE_ENV;
+const isProductionEnv = nodeEnv ? nodeEnv.startsWith('prod') : false;
+const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT;
+const isStagingEnv = objectivEnvironment ? (isProductionEnv && objectivEnvironment.startsWith('staging')) : false;
```
sdirosa, line 9: Very nice. I like how you check both environments in conjunction here.

We could also throw an error when staging is set on non-prod environments (unless we plan to have a dev-staging 🙄)

sdirosa:
@ivarpruijn I think this looks pretty good. 


I am wondering if you could replace external `Link`s with regular `<a>` tags with relative URLs (and waitUntilTracked them). Then you should be able to simply put relative URLs without the initial forward slash in the settings. Not sure if it works though

docusaurus.config.js
ivarpruijn: Good one. Think throwing the error is more of a nice-to-have though, as it doesn't interfere with any of the dev settings.

ivarpruijn:
> @ivarpruijn I think this looks pretty good.
> 
> I am wondering if you could replace external `Link`s with regular `<a>` tags with relative URLs (and waitUntilTracked them). Then you should be able to simply put relative URLs without the initial forward slash in the settings. Not sure if it works though

Interesting idea. And then set `target='_self'` for every `<a>` by default, unless specified otherwise. But I think that means a UX hit, because things will load slower. I think we should keep this in mind as a potential fix, but try the `trailingSlash` option first. WDYT?

docs/docusaurus.config.js
sdirosa: Oh right, that's not swizzled yet, isn't it.

Maybe we can just put this bit in variable and reuse it? This way we don't have the ternary duplicated all over the place.

```
const websiteURL = isStagingEnv ? 'https://staging.objectiv.io/' : 'https://objectiv.io/';
```

And below you can:

```
{
     label: 'Objectiv.io \u{1F855}',
     position: 'right',
     to: websiteURL,
     target: '_self',
}
```

```
items: [
     {
          label: 'Privacy Policy',
          to: websiteURL + 'privacy',
     },
     {
          label: 'Cookies',
          to: websiteURL + 'privacy/cookies',
     },
],
```

sdirosa:
> Interesting idea. And then set `target='_self'` for every `<a>` by default, unless specified otherwise. But I think that means a UX hit, because things will load slower. I think we should keep this in mind as a potential fix, but try the `trailingSlash` option first. WDYT?

I agree. Let's fix the slashes first. 

I've left a suggestion to keep things a bit cleaner, without too many urls around, in the thread above.

docs/docusaurus.config.js
ivarpruijn: Will do. Tried that at first but failed, but just thought of how to do it.

docusaurus.config.js
sdirosa: Sure. I was more aiming at notifying the dev of an impossible set of configuration options

docs/docusaurus.config.js
ivarpruijn: Did you have something like this in mind? We could maybe also abort the start/build. https://github.com/objectiv/objectiv.io/pull/120/commits/4b86d1f2c18094d6f8bd42fdb076397b3d0da4cd

sdirosa: Yes, although the condition should be `if (isProductionEnv` and not `if (!isProductionEnv` without the negation, right?

sdirosa: Oh wait, nvm, you negated both :)

ivarpruijn: Great. And see I posted in the wrong thread, but anyway :)

PR #121, opened 2021-11-16
Fix logo margin and center image on taxonomy intro

ivarpruijn:
- Logo doesn't have the erroneous 20px top margin.
- Image in the Taxonomy intro is centered.


jansenbob (approved):
LGTM if the env variable is fixed as mentioned

PR #122, opened 2021-11-16
Fix start error on null OBJECTIV_ENVIRONMENT value

ivarpruijn:
Fixes a startup error on the site & docs when the `OBJECTIV_ENVIRONMENT` is not set in the environment variables.

Thanks for the find @jansenbob


docs/docusaurus.config.js
```diff
@@ -5,7 +5,7 @@ const path = require('path');
 
 const nodeEnv = process.env.NODE_ENV;
 const isProductionEnv = nodeEnv ? nodeEnv.startsWith('prod') : false;
-const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT;
+const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT ?? 'dev';
```
sdirosa, line 8: I think we should default to `nodeEnv`. Else a production build without `OBJECTIV_ENVIRONMENT` set will internally set the `objectivEnvironment` to 'dev`. 

While the only allowed value for `OBJECTIV_ENVIRONMENT` is 'staging', right?

docusaurus.config.js
```diff
@@ -3,10 +3,16 @@
...
-const isStagingEnv = objectivEnvironment ? (isProductionEnv && objectivEnvironment.startsWith('staging')) : false;
+const nodeEnv = process.env.NODE_ENV;
+// Ensure that production builds are not run on local machines, and vice versa
+if (nodeEnv != objectivEnvironment) {
+  throw new Error("WARNING: NODE_ENV ("+nodeEnv+") and OBJECTIV_ENVIRONMENT (" + objectivEnvironment + ") "
+    + "are not equal");
+}
+const isProductionEnv = objectivEnvironment ? objectivEnvironment.startsWith('prod') : false;
+const isStagingEnv = objectivEnvironment ? (objectivEnvironment.startsWith('staging')) : false;
+const websiteUrl = isStagingEnv ? 'https://staging.objectiv.io/' : 'https://objectiv.io/';
```
sdirosa, line 15: Can we do this exact same thing also for `baseUrl`, `trackerDocsApplicationId` and `trackerEndPoint`.

Eg:

```
const baseUrl = isProductionEnv ? '/docs/' : '/';
const trackerDocsApplicationId = `objectiv-docs${isProductionEnv ? '' : isStagingEnv ? '-staging' : '-dev'}`;
const trackerEndPoint = isProductionEnv ? 'https://collector.objectiv.io' : 'http://localhost:5000';
```

This way we the config is clean from conditionals and in the future we can move this logic out in a shared module perhaps.

docs/docusaurus.config.js
```diff
@@ -3,15 +3,16 @@
...
 const nodeEnv = process.env.NODE_ENV;
-const isProductionEnv = nodeEnv ? nodeEnv.startsWith('prod') : false;
-const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT;
-const isStagingEnv = objectivEnvironment ? (isProductionEnv && objectivEnvironment.startsWith('staging')) : false;
-const websiteUrl = isStagingEnv ? 'https://staging.objectiv.io/' : 'https://objectiv.io/';
-if (!isProductionEnv && !objectivEnvironment.startsWith('dev')) {
-  console.error("WARNING: Running in development mode, but OBJECTIV_ENVIRONMENT is set to '" 
-    + objectivEnvironment + "'")
+// Ensure that production builds are not run on local machines, and vice versa
+if (nodeEnv != objectivEnvironment) {
```
sdirosa, line 9:
```suggestion
if (nodeEnv !== objectivEnvironment) {
```

docusaurus.config.js
```diff
@@ -3,10 +3,16 @@
 
 const path = require('path');
 
+const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT ?? 'development';
 const nodeEnv = process.env.NODE_ENV;
-const isProductionEnv = nodeEnv ? nodeEnv.startsWith('prod') : false;
-const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT;
-const isStagingEnv = objectivEnvironment ? (isProductionEnv && objectivEnvironment.startsWith('staging')) : false;
+// Ensure that production builds are not run on local machines, and vice versa
+if (nodeEnv != objectivEnvironment) {
```
sdirosa, line 9:
```suggestion
if (nodeEnv !== objectivEnvironment) {
```

docusaurus.config.js
```diff
@@ -16,7 +22,7 @@ const config = {
   titleDelimiter: '|',
   tagline: 'A data collection & modeling library that puts the data scientist first.', //meta description, and og:description
   baseUrl: '/',
-  url: isStagingEnv ? 'https://staging.objectiv.io/' : 'https://objectiv.io/',
+  url: websiteUrl,
```
sdirosa, line 25: Very nice, see [my comment](https://github.com/objectiv/objectiv.io/pull/122/files#r753133586) above about doing the same for all of these conditionals

sdirosa:
@ivarpruijn I've left a few suggestions

docusaurus.config.js
ivarpruijn: That's a great idea. Will update.

docusaurus.config.js
ivarpruijn: Will do

docusaurus.config.js
sdirosa: Very nice!

docusaurus.config.js
sdirosa: I think it's much cleaner, and easier to reconfigure, like this

sdirosa:
No more comments. Looks good!

PR #123, opened 2021-11-16
WIP: Automatically open first child of a category

sdirosa:
@borft FYI

### DIFF
--- a/docs/src/theme/DocSidebarItem/index.tsx
+++ b/docs/src/theme/DocSidebarItem/index.tsx
@@ -148,6 +148,9 @@ function DocSidebarItemCategory({item, onItemClick, activePath, ...props}) {
         onClick={
           collapsible
             ? (e) => {
+                if(!collapsed) {
+                  e.preventDefault();
+                }
                 toggleCollapsed();
                 trackClick({ element: e.target })
               }
--- a/docs/src/theme/DocSidebarItem/index.tsx
+++ b/docs/src/theme/DocSidebarItem/index.tsx
@@ -116,7 +116,7 @@ function DocSidebarItemCategory({item, onItemClick, activePath, ...props}) {
 
       // Else keep searching deeper
       case 'category':
-        findFirstLinkItem(firstItem.items)
+        return findFirstLinkItem(firstItem.items)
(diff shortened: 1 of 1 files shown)


docs/src/theme/DocSidebarItem/index.tsx
```diff
@@ -126,16 +129,15 @@ function DocSidebarItemCategory({item, onItemClick, activePath, ...props}) {
...
           collapsible
             ? (e) => {
-                e.preventDefault();
                 toggleCollapsed();
                 trackClick({ element: e.target })
               }
             : undefined
         }
-        href={collapsible ? '#' : undefined}
+        href={collapsible ? firstItem.href : undefined}
```
sdirosa, line 137: @borft I think we need some more logic here to not execute this link when collapsing. Eg: in the onClick event above we have to execute the e.preventDefault if the user clicks on the same category that's already in the current route, and skip it in all other cases. I'll fix it up tomorrow (I believe there's a helper API from Docusaurus to check this exact case already)

docs/src/theme/DocSidebarItem/index.tsx
```diff
@@ -100,6 +100,9 @@ function DocSidebarItemCategory({item, onItemClick, activePath, ...props}) {
     collapsed,
     setCollapsed,
   });
+
+  const [firstItem] = items;
+
```
sdirosa, line 105: If we have categories in categories without a page in between this won't work and we should write instead a recursive function to find the first child actual link

sdirosa: Fixed in latest version

docs/src/theme/DocSidebarItem/index.tsx
sdirosa: Fixed in latest version

docs/src/theme/DocSidebarItem/index.tsx
sdirosa: Fixed again, the `category` case was broken, couldn't really test it because I don't know how to generate the docs and I had the old version without nested categories

PR #124, opened 2021-11-17
Swizzle DocSidebar fully to fix its scrolling behavior

ivarpruijn:
Issue: the sidebar lost its sticky behavior. 

Problem: we added tracking by wrapping the sidebar in another div, which interfered with the `position: sticky` CSS applied to the specific order of sidebar/content containers.

Solution in this PR: swizzle the DocSidebar component, remove the wrapper for tracking, and add tracking inline.


sdirosa (approved):
Looks good, but I think this could have been fixed with tagChildren from a higher component. Let's try to do that later!

PR #126, opened 2021-11-17
updating LocationStack explanation

vincenthoogsteder:
better explaining that LS is a hierarchy of UI as the user sees it

### DIFF
--- a/docs/docs/taxonomy/reference/location-contexts/overview.md
+++ b/docs/docs/taxonomy/reference/location-contexts/overview.md
@@ -7,7 +7,7 @@ title: Overview
 import Mermaid from '@theme/Mermaid';
 
 LocationContexts are used to populate a [Tracker's](/tracking/core-concepts/trackers) or 
-[Event's](/tracking/core-concepts/events.md) `location_stack` properties.  A Location Stack is a visual representation of the UI location where a user is when an [Event](/tracking/core-concepts/events.md) occurs.  It is composed of a hierarchical stack of UI elements, meaning that the 
+[Event's](/tracking/core-concepts/events.md) `location_stack` properties.  A Location Stack is a visual representation of the UI location a user is currently viewing when an [Event](/tracking/core-concepts/events.md) occurs.  It is composed of a hierarchical stack of UI elements, meaning that the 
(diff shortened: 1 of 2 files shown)


docs/docs/taxonomy/reference/location-contexts/overview.md
```diff
@@ -7,9 +7,8 @@ title: Overview
 import Mermaid from '@theme/Mermaid';
 
 LocationContexts are used to populate a [Tracker's](/tracking/core-concepts/trackers) or 
-[Event's](/tracking/core-concepts/events.md) `location_stack` properties. A Location Stack is 
-meant to describe accurately where an [Event](/tracking/core-concepts/events.md) originated in the UI, e.g. 
-Sections, Menus, etc.
+[Event's](/tracking/core-concepts/events.md) `location_stack` properties.  A Location Stack is a visual representation of the UI location where a user is when an [Event](/tracking/core-concepts/events.md) occurs.  It is composed of a hierarchical stack of UI elements, meaning that the 
+order in the stack is the order in the UI as well.
```
ivarpruijn, line 11: Nice. Small suggestion (for both pages):
```suggestion
[Event's](/tracking/core-concepts/events.md) `location_stack` properties.  A Location Stack is a visual representation of the UI location for a user when an [Event](/tracking/core-concepts/events.md) occurs. It is composed of a hierarchical stack of UI elements, meaning that the order in the stack is the order in the UI as well.
````

vincenthoogsteder: updated

ivarpruijn (approved):

docs/docs/taxonomy/reference/location-contexts/overview.md
jansenbob: I find this confusing. I think the word _visual_ is in the wrong place. A location stack is not a visual representation imho. That sounds like it is an image or diagram of some sort. 

A location stack represents from which exact visual UI location an event occurs.

vincenthoogsteder:
I like that

jansenbob:
Decided to tackle this once the taxonomy update is in place

PR #128, opened 2021-11-17
Fix 404 in Modeling for 'Suggest an edit' in footer

ivarpruijn:
The baseUrl wasn't incorporated, so on production /docs/modeling wasn't detected. This PR should fix that, let's test on staging.


docs/src/theme/EditThisPage/index.tsx
```diff
@@ -16,10 +16,10 @@ import { tagLink, tagElement } from "@objectiv-analytics/tracker-browser";
 
 export default function EditThisPage({editUrl}: Props): JSX.Element {
   const context = useDocusaurusContext();
-  const {customFields} = context.siteConfig;
+  const {baseUrl, customFields} = context.siteConfig;
 
   // if in the 'modeling' section, edit the source code that is used to auto-generate the docs instead
-  let editThisPageUrl = useRouteMatch("/modeling") !== null ? 
+  let editThisPageUrl = useRouteMatch(baseUrl + "modeling") !== null ?
```
borft, line 22:
```suggestion
  let editThisPageUrl = useRouteMatch(`${baseUr}modeling`) !== null ? 
```

borft (approved):
lgtm, small suggestion about string composition, rather than concatination

ivarpruijn:
> lgtm, small suggestion about string composition, rather than concatination

Thanks, updated. Will deploy to staging for final check.

PR #145, opened 2021-11-22
Disable blog

ivarpruijn:
We see some tracking data for the blog, which is currently empty, most likely because it gets crawled. This PR disables the blog for now.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -45,6 +45,7 @@ const config = {
           customCss: require.resolve('./src/css/custom.css'),
         },
         docs: false,
+        blog: false,
         sitemap: {
           changefreq: 'weekly',
           priority: 0.5,
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -45,7 +45,10 @@ const config = {
           customCss: require.resolve('./src/css/custom.css'),
         },
         docs: false,
-        blog: false
+        sitemap: {
+          changefreq: 'weekly',
+          priority: 0.5,
+        },
       }),
     ],
   ],
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -45,18 +45,7 @@ const config = {
           customCss: require.resolve('./src/css/custom.css'),
         },
         docs: false,
-        blog: {
-          blogTitle: 'Objectiv Blog',
-          blogDescription: 'Objectiv Blog',
-          blogSidebarCount: 0,
-          postsPerPage: 5,
-          showReadingTime: false,
-          editUrl: 'https://github.com/facebook/objectiv/objectiv.io/edit/master/blog/',
-        },
-        sitemap: {
-          changefreq: 'weekly',
-          priority: 0.5,
-        },
+        blog: false
       }),
     ],
   ],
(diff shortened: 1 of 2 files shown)


sdirosa:
@ivarpruijn Shouldn't we create a robots.txt in the static folder to address this issue on a more general level?

ivarpruijn:
> @ivarpruijn Shouldn't we create a [robots.txt](https://developers.google.com/search/docs/advanced/robots/create-robots-txt) in the static folder to address this issue on a more general level?

I think that's a good idea in general. Currently we have no content for the blog though (so you see a directory listing at https://objectiv.io/blog/), so I think we should disable the blog until we have the first content.

sdirosa:
> > @ivarpruijn Shouldn't we create a [robots.txt](https://developers.google.com/search/docs/advanced/robots/create-robots-txt) in the static folder to address this issue on a more general level?
> 
> I think that's a good idea in general. Currently we have no content for the blog though (so you see a directory listing at https://objectiv.io/blog/), so I think we should disable the blog until we have the first content.

Sure but that doesn't meant crawlers are not gonna try to access it, it will just cause them to get a 404.

Some crawlers, like Google or Bing, are quite aggressive and can detect common website frameworks (not sure if Docusaurus is among them, but it's likely) and attempt to crawl sections that are commonly used.

I'm okay if that gets done in another PR of course

sdirosa:
@ivarpruijn Another thing I noticed while testing this quickly, our meta description tags for the modeling sections contain some sort-of JS code (``` export const toc = [ ```), I don't think that's intended

ivarpruijn:
> 

Makes sense. This PR would also remove /blog from our sitemap.xml, but if crawlers detect common frameworks and try to crawl it anyway, it makes sense to add a robots.txt. I'll add it to this PR!

ivarpruijn:
> @ivarpruijn Another thing I noticed while testing this quickly, our meta description tags for the modeling sections contain some sort-of JS code (`export const toc = [`), I don't think that's intended

That doesn't look good, I'll look into it, thanks.

ivarpruijn:
> > > @ivarpruijn Shouldn't we create a [robots.txt](https://developers.google.com/search/docs/advanced/robots/create-robots-txt) in the static folder to address this issue on a more general level?
> > 
> > 
> > I think that's a good idea in general. Currently we have no content for the blog though (so you see a directory listing at https://objectiv.io/blog/), so I think we should disable the blog until we have the first content.
> 
> Sure but that doesn't mean crawlers are not gonna try to access it, it will just cause them to get a 404.
> 
> Some crawlers, like Google or Bing, are quite aggressive and can detect common website frameworks (not sure if Docusaurus is among them, but it's likely) and attempt to crawl sections that are commonly used.
> 
> I'm okay if that gets done in another PR of course

@sdirosa This adds the robots.txt file, which gets included in a build: https://github.com/objectiv/objectiv.io/pull/145/commits/85a521ea8a244c6629defc32d7c71bd435536336

sdirosa:
> @sdirosa This adds the robots.txt file, which gets included in a build: [85a521e](https://github.com/objectiv/objectiv.io/commit/85a521ea8a244c6629defc32d7c71bd435536336)

I think this should work 👌🏼

I see that the sitemap contains the blog archive, but the robots are correct. So we shouldn't get any more crawling attempts, nor 404s

sdirosa (approved):

PR #147, opened 2021-11-22
WIP: fix environment logic

borft:
An attempt to simplify environment specific settings

I also don't like how cross-env blindly overrules environment variables (they are hard-coded). Perhaps we can find a more elegant solution for that.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -4,6 +4,12 @@
 const path = require('path');
 
 const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT ?? 'development';
+const nodeEnv = process.env.NODE_ENV;
+
+// only allow yarn start (dev) in dev mode
+if (nodeEnv === 'development' && objectivEnvironment !== 'development'){
+  throw new Error(`Not allowed to use 'yarn start' on non dev build (OBJECTIV_ENVIRONMENT=${process.env.OBJECTIV_ENVIRONMENT})`);
+}
 
 // this is development, which is the default
 let baseUrl = '/';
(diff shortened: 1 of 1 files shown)


sdirosa:
> An attempt to simplify environment specific settings
> 
> I also don't like how cross-env blindly overrules environment variables (they are hard-coded). Perhaps we can find a more elegant solution for that.

I quite like these changes. I would go even further and move the whole ENV logic out of the config in its own module.

About cross-env; maybe we don't even need it? We have a meeting tomorrow to discuss exactly these aspects. Let's hear also Ivar's thoughts.

Personally I don't have a very strong preference as I've never used nor I think I will ever need to use `yarn build`.  
As long as things are simple (eg: less dependencies, one configuration for both website and docs, etc) and safe (nobody should be able to, in any way, send localhost data to the production collector) I'm happy :)

docs/docusaurus.config.js
```diff
@@ -5,18 +5,33 @@ const path = require('path');
...
+  trackerConsoleEnabled = false;
+  websiteUrl = 'https://objectiv.io/';
+  trackerEndPoint = 'https://collector.objectiv.io';
+} else if ( objectivEnvironment === 'staging' ){
+  baseUrl = '/docs/';
+  trackerApplicationId = 'objectiv-docs-staging';
+  trackerConsoleEnabled = true;
+  websiteUrl = 'https://staging.objectiv.io/';
+  trackerEndPoint = 'https://collector.objectiv.io';
+}
```
sdirosa, line 33: This looks way more readable than it was before. I still don't understand why aren't all of these ENV variables instead of being hardcoded in here, but maybe there's some technical limitation (Windows? Docker?) of which I'm not aware.

ivarpruijn:
Closing as discussed, new proposal will follow

PR #168, opened 2021-12-01
Remove announcement bar

ivarpruijn:
- Removes the announcement bar
- Adds a bit of margin to the hero element, looks slightly better in my opinion now that the bar is gone. WDYT @jansenbob ?


sdirosa (approved):
Nice, and I agree that a bit more spacing looks nicer for the hero 👍🏼

jansenbob (approved):
Agree. Looks good. Ship it

PR #172, opened 2021-12-02
Style docs footer

ivarpruijn:
- Adds icons
- Styles it so it takes less attention


sdirosa (approved):
What an improvement! Really like this

jansenbob (approved):
very cool. LGTM

PR #173, opened 2021-12-03
Use dotenv for config

borft:


docs/docusaurus.config.js
```diff
@@ -3,19 +3,20 @@
 
 const path = require('path');
 
-const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT ?? 'development';
+const fs = require('fs')
```
borft, line 6: @sdirosa I'm now repeating these lines of code in both configs... On the one hand, it's pretty trivial, otoh, maybe we could package it in a separate file, and require it here? wdyt?

package.json
```diff
@@ -7,7 +7,8 @@
     "start": "docusaurus start",
     "start:docs": "cd docs && docusaurus start",
     "tsc": "tsc && (cd docs && tsc)",
-    "build": "cross-env OBJECTIV_ENVIRONMENT=dev docusaurus clear && docusaurus build",
```
borft, line 10: this was of course broken, it would completely ignore the env setting. So I've changed it to a plain build command that uses the environment variable, and added an explicit build:development for Ivar

docs/docusaurus.config.js
```diff
@@ -4,38 +4,27 @@
...
+  url: envConfig.websiteUrl,
+  baseUrl: envConfig.baseUrl,
   favicon: 'img/favicon/favicon.ico',
   organizationName: 'objectiv', // Usually your GitHub org/user name.
   projectName: 'objectiv.io', // Usually your repo name.
 
   onBrokenLinks: 'log',
   onBrokenMarkdownLinks: 'throw',
-  trailingSlash: false,
+  //trailingSlash: false,
```
borft, line 27: this is an experiment, so this options has 3 settings (true, false, undefined: https://docusaurus.io/docs/api/docusaurus-config#trailing-slash)

jansentom: how about making undefined explicit instead of implicit in that case? 
>> `trailingSlash: undefined,`

docs/docusaurus.config.js
sdirosa: Sounds like a good idea to me. We could just create a .js file somewhere high up and include it in both places

docs/docusaurus.config.js
borft: good point, I'll try that, add a comment as well (provided I can make it work)

sdirosa: +1 to make this explicit and maybe add the link to the docs above it. I do like the idea that Docusaurus doesn't fiddle with URLs with our current setup

docs/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "docusaurus": "docusaurus",
     "start": "docusaurus start",
-    "build": "cross-env OBJECTIV_ENVIRONMENT=dev docusaurus clear && docusaurus build",
+    "build": "docusaurus clear && docusaurus build",
+    "build:development": "cross-env OBJECTIV_ENVIRONMENT=dev docusaurus clear && docusaurus build",
```
sdirosa, line 9: I was wondering if we even need this one, wouldn't it be the same as `start`?
Is it meant to test if the build behaves the same as dev perhaps?

docusaurus.config.js
```diff
@@ -150,6 +138,11 @@ const config = {
   })
 };
 
+// only load the post-build plugin when creating a production build
+if ( objectivEnvironment === 'production' ){
+  config.plugins.push(path.resolve(__dirname, 'src/plugins/post-build/'));
+}
```
sdirosa, line 144: Another option, more in line with how Docusaurus plugins work, would be to leave the plugin always active and pass it a parameter to enable it or not. Eg:

```javascript
plugins: [
    path.resolve(__dirname, 'src/plugins/favicons/'),
    [
        postBuildPlugin,
        {
            skip: objectivEnvironment !== 'production'
        }
    ]
    require.resolve('docusaurus-plugin-image-zoom')
  ],
```

And, of course, in the Plugin itself :


```javascript
module.exports = function (context, { skip }) {
  return {
    name: 'add-htaccess',
    async postBuild({siteConfig = {}, routesPaths = [], outDir}) {
      if (skip) {
         console.log('Skipping post-build plugin');
         return;
      }
      // copy .htaccess from root to ./build
      // destination will be created or overwritten by default.
      fs.copyFile('.htaccess', path.join('build', '.htaccess'), (err) => {
        if (err) throw err;
        console.log('.htaccess file was copied to the build directory');
      });
    },
  };
};

```

docs/docusaurus.config.js
sdirosa: I see you already did ;) nice!

sdirosa (approved):
I've left some minor comments and one suggestion regarding the post-build plugin. For the rest, this is quite nice. No more dynamic configurations!

docs/package.json
borft: I'd say so. I personally don't care. But I think Ivar's OS of choice doesn't support environment variables

sdirosa: Cross-env should take care of that hopefully

sdirosa: I just tested on my VM, with powershell envs works fine. The syntax is a bit different ($env:OBJECTIV_ENVIRONMENT="dev") and I think that's why we have cross-env here. 


Found also a note that cross-platform compatibility works only when using `require('dotenv').config();` as explained [here](https://github.com/motdotla/dotenv#usage). A regular import doesn't seem to work on Windows. Go figure!

borft: well the thing is, cross-env doesn't actually use environment variables, which is why it's kind of useless, but I'm fine with keeping it for windows users ;)

sdirosa: Oh really? I assumed that to be its main purpose... 🤦🏼 

Then I'd remove it and make sure our custom loader works cross platform. 

For reference, this is the default loader I mentioned above: [require('dotenv').config();](https://github.com/motdotla/dotenv/blob/27dfd3f034ce00b1daa72effbd91dd7788aced48/lib/main.js#L82-L115)

The only change on our side is that we want to support .env.someting (dotenv doesn't do that by design), but the rest of the logic should be the same and we could replicate it as is.

.env.staging
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "website-staging"
+trackerEndPoint = "https://collector.objectiv.io"
```
thijs-obj, line 2: Do we not have a separate tracker for staging?

Makefile
```diff
@@ -0,0 +1,31 @@
...
+# build the docs, we check for index.html to determine if we need to do anything
+docs/build/index.html:
+	cd docs && yarn install && yarn build
+
+build-website: | clean-website build/index.html
+build-docs: | clean-docs docs/build/index.html
+
+# build docker container for full website, including docs
+# set environment to docker, to make sure the right config/env is loaded
+build-docker-website: OBJECTIV_ENVIRONMENT=docker
```
thijs-obj, line 28: Question: how does this work? Can you just define `build-docker-website` twice?

.env.staging
borft: no, we currently only have 1 collector

Makefile
borft: yes, the first one sets the variable, the second one defines what to do

thijs-obj: cool. didn't know you could write that like that

.env.staging
thijs-obj: ack

docs/src/components/sphinx-page/index.tsx
```diff
@@ -142,8 +143,22 @@ const SphinxPage = (props) => {
                     }
                     // only remove the .html if local links, leave external links alone
                     if ( isInternal ){
+                        console.log(`Found link to: ${a.href} --> ${currentSite} // ${window.location.toString()}`);
                         a.href = a.href.replace(/\.html/g, '');
 
+                        // try and fix links that are broken in builds do to slashes
+                        // we go from http://site/docs/modeling/page/otherpage/#anchor
+                        // to
+                        // http://site/docs/modeling/otherpage/#anchor
```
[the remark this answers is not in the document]
borft: The reason we have to do this, is the default trailingSlashes setting in docusaurus. It adds a level of nesting by moving `module/page.html` to `module/page/index.html`. This causes links in that page (in Sphinx pages) to be broken, if they're in a different module. eg module2/page2.html becomes module2/page2/index.html. However, if it is linked from the first page, the link would become module/module2/page2/index.html. In this case, we remove the first module/ from the URL to make it work again.

docs/docusaurus.config.js
```diff
@@ -4,38 +4,30 @@
...
+  url: envConfig.websiteUrl,
+  baseUrl: envConfig.baseUrl,
   favicon: 'img/favicon/favicon.ico',
   organizationName: 'objectiv', // Usually your GitHub org/user name.
   projectName: 'objectiv.io', // Usually your repo name.
 
   onBrokenLinks: 'log',
   onBrokenMarkdownLinks: 'throw',
-  trailingSlash: false,
+
```
borft, line 27: @ivarpruijn this is also an important change

docs/package.json
ivarpruijn, line 9: @borft I think this can go, and we need to add `dotenv` as a dependency.

docs/README.md
```diff
@@ -21,15 +21,24 @@ without having to restart the server.
...
+The configs for the environments are in their respectiv dotenv files. eg. production settings are in `.env.production`. 
+Both the website and docs (being separate Docusaurus instances) have their own set of config files.
+
+Then to build:
+
+```console
+yarn build
+```
+This command generates static content into the `build` directory and can be served using any static contents 
+hosting service.
```
ivarpruijn, line 41: @borft This seems duplicated (or the part below it is)

.env.development
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "website-dev"
```
ivarpruijn, line 1: If we want to keep backwards compatibility, this should be `objectiv-website-dev`.

.env.production
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "website"
```
ivarpruijn, line 1: If we want to keep backwards compatibility, this should be `objectiv-website`.

.env.staging
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "website-staging"
```
ivarpruijn, line 1: If we want to keep backwards compatibility, this should be `objectiv-website-staging`.

docs/.env.development
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "website-docs-dev"
```
ivarpruijn, line 1: If we want to keep backwards compatibility, this should be `objectiv-docs-dev`.

docs/.env.production
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "website-docs"
```
ivarpruijn, line 1: If we want to keep backwards compatibility, this should be `objectiv-docs`.

docs/.env.staging
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "website-docs-staging"
```
ivarpruijn, line 1: If we want to keep backwards compatibility, this should be `objectiv-docs-staging`.

docs/README.md
borft: indeed, seems a copy/paste error

docs/docusaurus.config.js
ivarpruijn: Thanks, bye trailingSlashes :)

docs/package.json
borft: done

docs/.env.development
```diff
@@ -0,0 +1,5 @@
+trackerApplicationId = "objectiv-website-docs-dev"
```
ivarpruijn: @borft This should actually be `objectiv-docs-dev`; same issue for the other .envs

ivarpruijn (approved):

PR #184, opened 2021-12-14
Add command to update pre-installs

jansenbob:
Add `docker compose pull` to demo guides to prevent issues for users who've already installed te demos before.


borft (approved):
lgtm

PR #195, opened 2021-12-19
Fix Algolia crawling http URLs instead of https

ivarpruijn:
The Algolia logs showed a lot of HTTP URLs being crawled instead of HTTPS, due to redirects in the docs. This change is an attempt to fix that with the .htaccess.

@borft Could you review? We should test this carefully on production.

### DIFF
--- a/docs/.htaccess
+++ b/docs/.htaccess
@@ -1,9 +1,18 @@
+# redirect to https
+RewriteEngine On
+RewriteCond %{HTTP_HOST} ^objectiv.io [NC]
+RewriteCond %{HTTP:X-Forwarded-Proto} !https
+RewriteCond %{HTTPS} off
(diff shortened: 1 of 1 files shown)


docs/.htaccess
```diff
@@ -1,9 +1,17 @@
+# redirect to https
+RewriteEngine On
+RewriteCond %{HTTP:X-Forwarded-Proto} !https
+RewriteCond %{HTTPS} off
+RewriteRule ^(.*)$ https://%{HTTP_HOST}/$1 [R,L]
```
borft, line 5: @ivarpruijn this is on staging now. should work on both environments

ivarpruijn: Great. Seems to work for me. Think we can merge and deploy this.

borft (approved):

PR #197, opened 2021-12-20
Fix tracker logging settings

ivarpruijn:
- Correctly uses the setting for enabling logging for the tracker.
- Removes spaces in the .env files, according to https://github.com/motdotla/dotenv#usage.


docs/docusaurus.config.js
```diff
@@ -73,7 +73,7 @@ const config = {
     trackerDocsApplicationId: envConfig.trackerApplicationId,
     trackerEndPoint: envConfig.trackerEndPoint,
     slackJoinLink: slackJoinLink,
-    trackerConsoleEnabled: envConfig.trackerConsoleEnabled
+    trackerConsoleEnabled: envConfig.trackerConsoleEnabled === '1'
```
ivarpruijn, line 76: @sdirosa We could of course also simply check if the setting is there in the .env file, and then enable the logging.

docusaurus.config.js
```diff
@@ -67,7 +67,7 @@ const config = {
     trackerApplicationId: envConfig.trackerApplicationId,
     trackerEndPoint: envConfig.trackerEndPoint,
     slackJoinLink: slackJoinLink,
-    trackerConsoleEnabled: envConfig.trackerConsoleEnabled
+    trackerConsoleEnabled: envConfig.trackerConsoleEnabled === '1'
```
ivarpruijn, line 70: Same here as above

.env.development
```diff
@@ -1,5 +1,5 @@
-trackerApplicationId = "objectiv-website-dev"
-trackerEndPoint = "http://localhost:5000"
-websiteUrl = "http://localhost:3000"
-baseUrl = "/"
-trackerConsoleEnabled = true
+trackerApplicationId="objectiv-website-dev"
+trackerEndPoint="http://localhost:5000"
+websiteUrl="http://localhost:3000"
+baseUrl="/"
+trackerConsoleEnabled=1
```
borft, line 5: works for me. As we're matching string, we can also still keep the true/false ofc. Just need to make sure we match for strings (eg. `==='true'`)

.env.docker
```diff
@@ -1,5 +1,5 @@
-trackerApplicationId = "objectiv-website-docker"
-trackerEndPoint = "http://localhost:5000"
-websiteUrl = "http://localhost:8080"
-baseUrl = "/"
-trackerConsoleEnabled = true
+trackerApplicationId="objectiv-website-docker"
+trackerEndPoint="http://localhost:5000"
+websiteUrl="http://localhost:8080"
+baseUrl="/"
+trackerConsoleEnabled=1
```
sdirosa, line 5: This format will be soon deprecated (it's already in strict mode, which we don't have enabled), the correct one is:
```
trackerConsoleEnabled="1"
```

docs/docusaurus.config.js
sdirosa: This change is not necessary.
Since you decided to use '0' and '1', those already parse to truthy and falsy

ivarpruijn: If I remove this check, set `trackerConsoleEnabled="0"` in `.env.development`, and `yarn start` the docs, the tracker still does logging.

sdirosa: In developer mode the console is always enabled, it can only be overridden with a custom one, but not disabled. Without the Console you cannot effectively instrument the Tracker correctly because you would be missing on the validation checks.

sdirosa: This is the relevant code if you are interested: https://github.com/objectiv/objectiv-analytics/blob/main/tracker/trackers/browser/src/BrowserTracker.ts#L58-L60

ivarpruijn: Ah okay, that makes sense then.

However, just tested with `trackerConsoleEnabled="0"` in `.env.staging` and creating a staging build, and logging is still enabled when I `npm run docusaurus serve` it.

sdirosa: There's something odd going on with how envs are loaded. 

If I do a `yarn build:staging` this is what I see in console:

```
OBJECTIV TRACKER APPLICATION ID: objectiv-docs-dev
OBJECTIV TRACKER ENDPOINT: http://localhost:5000
OBJECTIV TRACKER CONSOLE ENABLED: true
DOCUSAURUS URL: /
DOCUSAURUS BASEURL: /
```

This looks like the dev env, not the staging one.

Do you have the same?

borft: I don't know. I always set the environment and use `yarn build`. Works fine for me.

`yarn build:staging` also doesn't work for me, it simply uses the `OBJECTIV_ENVIRONMENT` that's already set, rather then setting it itself.

sdirosa: Not that I can tell.

Regardless, seems like the cross-env project has been killed: https://github.com/kentcdodds/cross-env/issues/257
So I can't check if this is a known issue under Linux. 

Nonetheless, the current build:* commands do not work and actually are a bit dangerous imho.

I think we should switch to an alternative that works as intended or remove them

borft: How about:
```make OBJECTIVE_ENVIRONMENT=staging build-docs```

This is what I use

sdirosa: If that works on Windows I think we should be good

borft (approved):
lgtm

sdirosa:
Very nice cleanup 👌🏼

PR #198, opened 2021-12-20
Upgrade Mermaid to 8.13.6

ivarpruijn:
Minor upgrade with seemingly no security fixes, we could skip it. 

See release notes:
- [Mermaid](https://github.com/mermaid-js/mermaid/releases/tag/8.13.6).
- [DOMPurify](https://github.com/cure53/DOMPurify/releases/tag/2.3.4).

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -25,7 +25,7 @@
     "cross-env": "^7.0.3",
     "dotenv": "^10.0.0",
     "file-loader": "^6.2.0",
-    "mermaid": "^8.13.5",
+    "mermaid": "^8.13.6",
     "prism-react-renderer": "^1.2.1",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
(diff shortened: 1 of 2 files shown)


sdirosa (approved):
Looks good!

sdirosa:
Question out of curiosity: shouldn't Dependabot create PRs for these updates automatically? Or did we disable that?

ivarpruijn:
> Question out of curiosity: shouldn't Dependabot create PRs for these updates automatically? Or did we disable that?

I think we only enabled Dependabot to do this for security updates/alerts in the Settings. If we want automatic version updates we  should configure that in a `.github/dependabot.yml` config file I think: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/enabling-and-disabling-dependabot-version-updates

sdirosa:
> > Question out of curiosity: shouldn't Dependabot create PRs for these updates automatically? Or did we disable that?
> 
> I think we only enabled Dependabot to do this for security updates/alerts in the Settings. If we want automatic version updates we should configure that in a `.github/dependabot.yml` config file I think: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/enabling-and-disabling-dependabot-version-updates

Interesting, I didn't know the two options where in different places. Thanks for clarifying

PR #200, opened 2021-12-23
fix links in reference

borft:


borft:
Merging this, needed to make links work on prod

PR #203, opened 2021-12-23
Upgrade Mermaid to 8.13.7

ivarpruijn:
Security release to fix vulnerability with links from actors in sequence diagrams, see [release notes](https://github.com/mermaid-js/mermaid/releases/tag/8.13.7).

Tested briefly, and all diagrams seem to work.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -22,7 +22,7 @@
     "clsx": "^1.1.1",
     "dotenv": "^10.0.0",
     "file-loader": "^6.2.0",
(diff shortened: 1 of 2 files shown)


sdirosa:
I find it funny that their dependabot is bumping their dependencies to fix the issue, but our dependabot doesn't propose this update. Perhaps because it's flagged as pre-release

Let's merge it!

sdirosa (approved):

PR #212, opened 2022-01-03
Upgrade to Docusaurus beta 14

ivarpruijn:
- Upgrades Docusaurus to beta 14, both website & docs.
- Fixes one breaking change in the swizzled DocPaginator plugin.

Tested & built for both site & docs, compared to production, no issues found.


sdirosa (approved):
While testing I noticed that the lockfile for the website was out of date, I've pushed it.

Apart from that, this looks good to 🚢

PR #213, opened 2022-01-04
Fix active state styling for Taxonomy diagrams

ivarpruijn:
I noticed the active states for elements in the Taxonomy diagrams got a bit lost here and there, so I updated them.

@jansenbob came up with a style for the new contexts/events that cannot be clicked.


ivarpruijn:
Closing after discussing this offline

PR #216, opened 2022-01-05
Website: Upgrade to react tracker

sdirosa:
This PR swaps Browser Tracker with the experimental React Tracker. 

So far I haven't encountered any issues. The goal is to get to a version we are happy with, including some Location simplifications, and then test on staging thoroughly.

I also introduced some Wrapper Components that automate most of the tracking for the semantics the websites uses. These components, at least the more generic ones, may move to the React Tracker eventually.


.yarnrc
```diff
@@ -0,0 +1 @@
+"@objectiv-analytics:registry" "http://127.0.0.1:4873"
```
sdirosa, line 1: Temporary to test packages locally via Verdaccio

.yarnrc
```diff
@@ -0,0 +1 @@
+"@objectiv:registry" "http://127.0.0.1:4873"
```
sdirosa, line 1: This is temporary. We can remove it when the new version of the Tracker packages is on NPM.

docusaurus.config.js
```diff
@@ -99,31 +99,36 @@ const config = {
         {
           label: 'Docs',
           to: envConfig.websiteUrl + '/docs', // ensure Docusaurus redirects to standalone docs
-          target: '_self'
+          target: '_self',
+          external: true
```
sdirosa, line 103: The external bit tells the TrackedLink component to wait. This makes sure we will track press events when navigating to the Docs.

src/components/announcement-bar/index.tsx
```diff
@@ -1,35 +1,22 @@
...
-      <span className={styles.announcementStar}><img src="/img/icons/icon-star.svg" alt='star' /></span>
-    </div>
+    <TrackedDiv id={'announcement-bar'}>
+      <div className={clsx(styles.announcement)}>
+        <span className={styles.announcementStar}><img src="/img/icons/icon-star.svg" alt='star' /></span>
+          <strong>{title}</strong>&nbsp;
+          {content}&nbsp;
+          <TrackedLink to={ctaLink} external={true}>
+            {ctaText}
+          </TrackedLink>
```
sdirosa, line 16: Here another example of a link marked as external

src/pages/about.tsx
```diff
@@ -80,7 +59,7 @@ export default function AboutUs() {
   const {siteConfig} = context;
 
   return (
-    <div {...tagRootLocation({id: 'page-about'})}>
+    <TrackedRootDiv id={'page-about'}>
```
sdirosa, line 62: TrackedRootDiv is just a tiny wrapper that makes it easier to wrap its children in RootLocationContext

src/pages/about.tsx
```diff
@@ -111,16 +87,10 @@ export default function AboutUs() {
               process was tedious and inefficient, so we started looking for better ways.
             </p>
           </div>
-        </header>
+        </TrackedHeader>
```
sdirosa, line 90: I don't think this header does anything, probably a leftover, since it doesn't contain any interactions.

I think these cases may be interesting for future validation checks. Some situations, like this one, may be ambiguous. 

Is the div void of interactions because of a copy paste mistake, or a refactoring, or because we plan to add interactions to it in the future?

Hard, for us, to decide whether this is correct or not.

src/pages/about.tsx
```diff
@@ -206,16 +173,12 @@ export default function AboutUs() {
                 <div>
                   <strong>We have the right backing</strong><br />
                   We're backed by&nbsp;
-                  <Link 
-                    to="https://www.fly.vc/"
-                    {...tagLink({ id: 'vc-fly', href: 'https://www.fly.vc/' })}>
+                  <TrackedLink to="https://www.fly.vc/">
```
sdirosa, line 176: I did not add a wait to these links because they automatically open in new Tabs. This may break if Docusaurus changes this behavior in the future.

Perhaps it would be best to give them an `external={true}`

src/pages/how-it-works.tsx
```diff
@@ -1,46 +1,44 @@
...
+        <TrackedHeader>
           <div className={clsx('container', styles.intro)}>
             <h2>How it works</h2>
             <p>
               With Objectiv, you take, stack and run pre-built data models off the shelf to quickly <br />
               build highly specific model stacks for in-depth product analysis and exploration.
             </p>
           </div>
-        </header>
+        </TrackedHeader>
```
sdirosa, line 32: Another header without interactions

src/pages/how-it-works.tsx
```diff
@@ -1,46 +1,44 @@
...
 
-        <main 
-          {...tagContent({id: 'main'})}
-          className={clsx('body-large')}>
+        <TrackedMain className={clsx('body-large')}>
```
sdirosa, line 34: I think this main doesn't add much, as there's nothing between the root location and this location. To be double checked

src/pages/index.tsx
```diff
@@ -159,53 +145,42 @@ export default function Home() {
...
-                    target="_self" 
-                    className={clsx("button", styles.ctaButton)}>
+                    external={true}
+                    target="_self"
+                    className={clsx("button", styles.ctaButton)}
+                  >
                     <span><img src={useBaseUrl("img/icons/icon-docs-blue.svg")}  alt={'Objectiv Quickstart Guide'}/></span>
                     Objectiv Quickstart Guide
-                  </Link>
+                  </TrackedLink>
```
sdirosa, line 156: I added the external link to this URL, since it resides in the docs and it was not tracked 100% of the time

src/pages/index.tsx
```diff
@@ -1,27 +1,27 @@
...
 import styles from './styles.module.css';
+import { TrackedLink } from '../trackedComponents/TrackedLink';
 
 export default function Home() {
   const context = useDocusaurusContext();
   const {url, tagline, customFields} = context.siteConfig;
 
   return (
-    <div {...tagRootLocation({id: 'page-home'})}>
+    <TrackedRootDiv id={'page-home'}>
```
ivarpruijn, line 18: @sdirosa We set this `id` manually now. Do we plan on automating this?

src/theme/Logo/index.tsx
```diff
@@ -9,18 +9,15 @@ import React from 'react';
...
   const {
     navbar: {logo = {src: ''}},
   } = useThemeConfig();
   const logoLink = useBaseUrl(logo.href || '/');
 
-  return <OriginalLogo
-    {...tagLink({ id: 'logo', href: logoLink })}
-    {...props} 
-  />
+  return <TrackedLink Component={OriginalLogo} id={'logo'} href={logoLink} title={logo.alt} {...props} />
```
sdirosa, line 20: This is an interesting case. The original logo is a special type of Link that does not have href and title, it just redirects to the root by itself. Here we use it as Component of a TrackedLink and specify href and title ourselves to close the loop. 

This can be interpreted as a Tracked Link rendering as OriginalLogo.

What happens internally is that we enrich OriginalLogo with a custom onClick event triggering our PressEvent tracker.

src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -7,24 +7,21 @@
...
+    <TrackedLink
+      Component={OriginalDefaultNavbarItem}
       {...props}
-      {...tagLink({ 
-          id: props.label, 
-          href: props.href ? props.href : props.to,
-        })
-      }
+      title={props.title ?? props.label}
     />
```
sdirosa, line 25: Similar to the Logo, here we render a Tracked Link as OriginalDefaultNavbarItem.

We make sure to forward all props to it and the only tweak we had to apply is to the `title` property of the Link. NavbarItems use `label`s for some elements and we have to grab that value if a title is not available.

src/theme/Root.tsx
```diff
@@ -43,45 +45,21 @@ function Root({children}) {
...
-      }
-    }
-  }
+  // Create React Tracker instance
+  const tracker = new ReactTracker({
+    endpoint: trackerEndPoint as string,
+    console: trackerConsoleEnabled && windowExists() ? console : undefined,
+    applicationId: trackerApplicationId as string,
+    active: cookiebotStatisticsConsent,
+  })
```
sdirosa, line 54: React Tracker is safe to instantiate in SSR. Only interesting tweak here is the windowExists check. This prevents a bunch of logging to be printed out while building.

src/theme/Root.tsx
```diff
@@ -43,45 +45,21 @@ function Root({children}) {
...
-      }
-      if(cookiebotStatisticsConsent) {
-        getTrackerRepository().activateAll();
-      } else {
-        getTrackerRepository().deactivateAll();
-      }
-    },
-    [cookiebotStatisticsConsent] // execute every time `cookiebotStatisticsConsent` changes
-  )
+  useOnChange(cookiebotStatisticsConsent, () => tracker.setActive(cookiebotStatisticsConsent))
```
sdirosa, line 57: React Tracker doesn't need a Repository. 

Trackers can be created anywhere and passed to components by using ObjectivProvider and  TrackerProvider.

If a branch of the application needs a different tracker instance we can simply re-wrap that section with another provider.

src/trackedComponents/README.md
```diff
@@ -0,0 +1,2 @@
+The components in here are experimental and may eventually be moved to the Tracker for React package, it's handier 
+to have them in here for prototyping and testing them.
```
sdirosa, line 2: I will have to move and package these components in the react package eventually, but it's quite a bit of work as I've to unit tests them as well. 

For now I'll leave them here.

src/pages/index.tsx
sdirosa: It's not possible in Docusaurus because there's no definition of routes and their names

sdirosa: In the Docs I used the URL as a source of info, the first slug is what I used to generate a RootLocationContext. I don't know if that works for the website as well though

sdirosa: Here is the plugin I mentioned:

https://github.com/objectiv/objectiv.io/blob/upgrade-to-react-tracker/docs/src/theme/RootLocationContextFromURLPlugin.ts

If you think this can work for the website and well I can simply move this into a module and use it here as well.

src/theme/Navbar/index.tsx
```diff
@@ -173,15 +165,16 @@ function NavbarMobileSidebar({sidebarShown, toggleSidebar}) {
...
           type="button"
+          title={'close'}
           className="clean-btn navbar-sidebar__close"
           onClick={toggleSidebar}>
           <IconClose
             color="var(--ifm-color-emphasis-600)"
             className={styles.navbarSidebarCloseSvg}
           />
-        </button>
+        </TrackedButton>
```
sdirosa, line 177: I've just added TrackedPressable and TrackedButton, this completes all the interactive Elements.

TrackedPressable is a generic wrapper that accepts a component that can have onClick, TrackedButton is simply a shortcut to create a TrackedPressable with a <button> configured as component.

src/theme/Navbar/index.tsx
```diff
@@ -149,17 +151,7 @@ function NavbarMobileSidebar({sidebarShown, toggleSidebar}) {
...
-        id: 'hamburger-menu',
-        options: {
-          trackVisibility: {
-            mode: 'manual',
-            isVisible: sidebarShown
-          }
-        }
-      })}
-    >
+    <TrackedOverlay id={'hamburger-menu'} isVisible={sidebarShown} className="navbar-sidebar">
```
sdirosa, line 154: An example of overlay. This will trigger visibility events automatically, based on the value of `isVisible` which is bound to the `sidebarShown` variable.

The Overlay visibility events will monitor changes in the isVisible and trigger when it toggles from hidden-to-visible or visible-to-hidden. 

The latter will only trigger if the overlay was actually known to have been visible before.

This last condition is quite important as it ensures no HiddenEvent is triggered when the component mounts as hidden. In the old BrowserTracker it was up to the developer to ensure this behavior, but it's now automatic.

src/theme/Navbar/index.tsx
```diff
@@ -242,15 +234,16 @@ function Navbar() {
...
+              title={'hamburger'}
               aria-label="Navigation bar toggle"
               className="navbar__toggle clean-btn"
               type="button"
               tabIndex={0}
               onClick={mobileSidebar.toggle}
               onKeyDown={mobileSidebar.toggle}>
               <IconMenu />
-            </button>
+            </TrackedButton>
```
sdirosa, line 246: Both this button and the closing one above don't have any text that we can detect. That's why I've added the `title` prop. Without it the Tracker will trigger an exception.

In the future we may refine this behavior. Perhaps we can simply log the issue and discard the tracking for this element, instead of throwing.

To be discussed.

src/pages/about.tsx
sdirosa: See [discussion below ](https://github.com/objectiv/objectiv.io/pull/216/files#r779632757) on why do we assign these contexts by hand.

src/pages/how-it-works.tsx
```diff
@@ -60,11 +58,12 @@ export default function Home() {
                     with effective modeling in mind.</p>
                 </div>
               </div>
-            </div>
+            </TrackedDiv>
```
sdirosa, line 61: These TrackedDiv components are just shortcuts for `<TrackedContent Component='div' ...`. Since div are widely used as container elements, it seemed nicer to have a shorter syntax for them. We can add and remove these shortcuts very easily.

src/pages/about.tsx
ivarpruijn: In the latest homepage version I removed the TrackedHeader, because there are no elements tracked in it, so I thought I'd keep it clean.

src/pages/about.tsx
ivarpruijn: I think fine for now, these are not critical to track

src/pages/index.tsx
ivarpruijn: Other than for the homepage, I think this would work well for all the other pages, which all have a path like `/jobs`.

src/pages/index.tsx
ivarpruijn: Makes sense

docusaurus.config.js
ivarpruijn: works like a charm

src/pages/how-it-works.tsx
ivarpruijn: This whole page is removed in the new site PR, but in general I have replaced all `<TrackedHeader>`s with a regular `<header>` if there are no tracked elements within it.

src/theme/Logo/index.tsx
ivarpruijn: nice solution I think

src/theme/Root.tsx
ivarpruijn: Is the `windowsExsists()` check something we can automate without the dev having to do it?

src/pages/about.tsx
sdirosa: I think it's a good idea. On mobile I think the website gets suspended while a new tab is open, which pretty much prevents the tracker to finish doing its job.

src/pages/how-it-works.tsx
sdirosa: Unless we expect to have interactions in these headers at some point in the future, I think this is the way to go.

src/pages/index.tsx
sdirosa: Indeed. For the homepage there's a special fallback called 'home', you can see that at this line:
https://github.com/objectiv/objectiv.io/blob/upgrade-to-react-tracker/docs/src/theme/RootLocationContextFromURLPlugin.ts#L32

Monday I'll move this plugin in its own package, so we can use it wherever we need.

src/theme/Logo/index.tsx
sdirosa: The alternative would have been to build a custom Logo component, but that seemed way too much work for something like this that is not reused anywhere else.

src/theme/Root.tsx
sdirosa: We already do for dev mode, but in this case we are passing the console based on `trackerConsoleEnabled` variable. This takes precedence on our internals.

If you prefer to rely on those you can get rid of the variable and avoid passing the `console` option. The tracker will automatically enable the console if it detects node in dev mode in a browser.

sdirosa: The problem here is that our env files don't take in account SSR. Docusaurus basically builds each page twice, once for the server and once for the client. We only provide configurations for the client side of things.

src/pages/about.tsx
ivarpruijn: Ah good reason to do it then

src/pages/how-it-works.tsx
ivarpruijn: agree

src/pages/index.tsx
ivarpruijn: sounds good!

src/theme/Logo/index.tsx
ivarpruijn: agree

src/pages/index.tsx
sdirosa: @ivarpruijn Here you go: https://github.com/objectiv/objectiv.io/pull/216/commits/c2698ec0ba2e164976efa3a06e49f8b378d18f3d

sdirosa: You probably need do a `yarn clear` after rebuilding the tracker packages and before building the website

ivarpruijn: Nice!!

src/pages/jobs.tsx
```diff
@@ -36,56 +36,37 @@ export default function Jobs() {
...
+                  target="_self"
+                >
+                  site
+                </TrackedLink>,&nbsp;
+                <TrackedLink
+                  to={url + "/docs/"}
+                  autoAddBaseUrl={true}
+                  waitUntilTracked={true}
+                  target="_self"
+                >
```
sdirosa, line 56: Removed the `useBaseURL` hook and used instead the `autoAddBaseUrl` feature of Docusarus Link components. Did the same everywhere else.

src/pages/privacy/cookies.tsx
```diff
@@ -0,0 +1,90 @@
...
+              href={'#'}
+               onClick={event => {
+                 event.preventDefault();
+                 // @ts-ignore
+                 Cookiebot.renew();
+                 return false;
+               }}
+            >
+              Renew or change your cookie consent here
+            </a>.
```
sdirosa, line 49: The previous syntax (`href="javascript: Cookiebot.renew()"`) is deprecated and will be not allowed in the next release of React.

I switched to React syntax to execute this API.

src/trackedComponents/TrackedLink.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+import Link, { LinkProps } from "@docusaurus/Link";
+import { TrackedLinkContext, TrackedLinkContextProps } from "@objectiv/tracker-react";
+import React from 'react';
+
+export type TrackedLinkProps = Omit<TrackedLinkContextProps, 'Component' | 'href'> & LinkProps;
+
+export const TrackedLink = React.forwardRef<HTMLAnchorElement, TrackedLinkProps>((props, ref) => (
+  <TrackedLinkContext Component={Link} {...props} href={props.href ?? props.to} ref={ref}/>
+))
```
sdirosa, line 13: This little guy is the only custom tracked component left in the website. It basically wraps around Docusaurus Link and automates supporting both `to` and `href` property.

src/pages/jobs.tsx
ivarpruijn: didn't know that existed, nice one

src/pages/privacy/cookies.tsx
ivarpruijn: okay, thanks

sdirosa:
@ivarpruijn Using the [latest tracker with validation](https://github.com/objectiv/objectiv-analytics/pull/376) I found only 1 collision (duplicated footer in the home page) and everything else seems okay. 

I think we could merge this branch as is. 

We only have to remember to remove the `.yarnrc` eventually.

docusaurus.config.js
sdirosa: This flag has been finalized to `waitUntilTracked` for consistency with Core Tracker and Browser Tracker

src/pages/about.tsx
sdirosa: This has been removed and Root locations are now tracked automatically via a newly created plugin

src/trackedComponents/README.md
sdirosa: This is done and in main

sdirosa:
Merging this as it's unlikely we will need to revert to using the Browser Tracker from the website, now that the React Tracker is almost ready for deployment

PR #220, opened 2022-01-10
removing apple m1 notifications

vincenthoogsteder:


borft (approved):
looks good to me.

We should make sure not to deploy this, before https://github.com/objectiv/objectiv-analytics/pull/365 is merged

sdirosa (approved):

PR #223, opened 2022-01-13
WIP: Add Loom video

ivarpruijn:
This adds a Loom video to the homepage (temp URL for now). Uses an iframe.

I also added a prototype Loom component that uses their [embedSDK](https://dev.loom.com/docs/embed-sdk/getting-started), but didn't get it to work quickly, so maybe @sdirosa would like to have a stab at it. If not, let's remove it.


src/components/loom-video/index.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import * as loom from '@loomhq/loom-embed';
+
+function LoomVideo({url, target}) {
+  return loom.textReplace(url=url, target=target);
+}
+
+export default LoomVideo;
```
ivarpruijn, line 8: This is a test, should remove if this doesn't work

sdirosa: Oh man. Perhaps I can try to render this with OEmbed, since they support it, and that should allow us to gain a bit more control.

I'll give it a shot tomorrow.

ivarpruijn: Cool, thanks. I'm going to make one more update to the design first, Bob had a different idea.

sdirosa: Ok 👌🏼

ivarpruijn:
Looks like this one can be closed

PR #230, opened 2022-01-18
Add ContentContext to Reusable Models in home page

sdirosa:
@jansenbob FYI: this is a tiny change that hopefully will not bother you. As mentioned by Vincent in chat the Reusable Models section in home page did not have any ContentContext around it.

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -146,7 +146,9 @@ export default function Home() {
             </div>
           </div>
 
-          <div className={clsx(styles.pageSection)}>
+          <TrackedDiv
+            id={'reusable-models'}
+            className={clsx(styles.pageSection)}>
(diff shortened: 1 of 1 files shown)


src/pages/index.tsx
```diff
@@ -207,7 +209,7 @@ export default function Home() {
                 </TrackedLink>
               </div>
             </div>
-          </div>
+          </TrackedDiv>
```
sdirosa, line 212: My 2 cents: I think what happened here is that this section did not have any links, then we added one and forgot to add a ContentContext around it.

Maybe a nice idea for a validation rule, or a warning, would be to ask developers to always have at least 1 ContentContext around Links. I think in general having links sitting directly under RootLocations is a bit odd.

PR #235, opened 2022-01-19
React Tracker docs + Tracking section overhaul

sdirosa:
This PR contains the new React Tracker docs and a major restructuring of the Tracking section to make space to multiple platforms.

For content review: https://staging.objectiv.io/docs/tracking/

Notable changes:
- Split Angular documentation out of Browser Tracker
- Moved Browser Tracker and Angular Tracker documentation in root
- Renamed Tracker to SDK
- Added React SDK to the root
- Pushed Core Concepts to the bottom of the sidebar and introductory docs
- Moved How-to-Guides from root to specific Trackers
- Moved Core Concepts that were specific to a Tracker in their correct folder
- Removed Core Concept pages that were redundant with the Taxonomy section (eg: events)

TODO:

- [x] Proofread everything
- [x] Check all examples to make sure they are for the correct section they are in
- [x] Check all Taxonomy names to make sure there are no outdated Location Stacks or context names
- [x] Double check for broken links
- [ ] Check the overall narrative


docs/docusaurus.config.js
```diff
@@ -22,7 +22,7 @@ const config = {
   organizationName: 'objectiv', // Usually your GitHub org/user name.
   projectName: 'objectiv.io', // Usually your repo name.
 
-  onBrokenLinks: 'log',
+  onBrokenLinks: 'throw',
   onBrokenMarkdownLinks: 'throw',
```
sdirosa: Only functional change in the whole branch. I'm not sure why was this set to log, it's not like we are gonna deploy broken links... are we? :)

jansentom: Log can be useful to collect all broken links at once, but I don't think we need that anymore?

@borft FYI

sdirosa: Good point, maybe we can make this variable dynamic depending on the env then.

We can leave it as `log` in dev and make into `throw` while building for prod

jansentom: That sounds good. WDYT @borft

sdirosa: I tested this real quick and the throw also builds a report of all broken links, which is printed to the console before aborting the build.

The difference seems to be that log allows you to click around anyway. Eg: you can dismiss the crashes.
The throw will not allow yarn start or yarn build to complete instead.

borft: Well, I don't mind the build/env breaking on broken links actually ;) This would prevent you from committing broken stuff (unless you commit untested stuff). 

So I'd be OK with throw, rather than log

sdirosa: Cool 👌🏼  

If anyone needs to do a heavy refactoring, they can always temporarily change this option on their local env of course.

PR #243, opened 2022-01-25
remove open model hub from main docs

jansenbob:
This removes the open model hub docs from the main Docs section. (does not affect the OMH in the Modeling section)


vincenthoogsteder:
looks good to me, let's pls double check if there will be broken links when doing this @borft

vincenthoogsteder (approved):
lgtm

jansenbob:
> looks good to me, let's pls double check if there will be broken links when doing this @borft

@vincenthoogsteder that's a good point. I have previously already updated the links that i knew of, but i'm not sure if i've missed any.

ivarpruijn:
Just ran the broken links checker and the docs themselves, plus did a search for any docs linking to the one removed, and all looks OK to me. I'd say we can merge this.

PR #249, opened 2022-01-27
Video player

ivarpruijn:
Adds:
- A component for a video player from various sources, as we're thinking about moving away from Loom embeds, as they set a lot of cookies (and for instance Vimeo doesn't if you specify a `dnt` parameter).
- Tracking for video play/pause/stop. 

Not yet done: tracking for any CTA clicks at the end of a video. I've contacted Vimeo if this is possible, as I couldn't find the option in their player SDK docs.

@sdirosa What do you think of this approach?

### DIFF
--- a/src/components/video-player/index.js
+++ b/src/components/video-player/index.js
@@ -13,9 +13,8 @@ import styles from "./styles.module.css";
 
 function VideoPlayer({
   url,
-  id = "video",
-  enableControls = true,
-  showPlayPauseButton = false,
+  id,
+  enableControls = true
 }) {
   const [isPlaying, setPlaying] = useState(false);
 
@@ -24,7 +23,7 @@ function VideoPlayer({
   }
 
   return (
-    <MediaPlayerContextWrapper id={id} className={styles.videoWrapper}>
+    <MediaPlayerContextWrapper id={id}>
       {(trackingContext) => (
         <>
           <ReactPlayer
@@ -35,12 +34,8 @@ function VideoPlayer({
             onStart={() => trackMediaStartEvent(trackingContext)}
             onPause={() => trackMediaPauseEvent(trackingContext)}
             onEnded={() => trackMediaStopEvent(trackingContext)}
+            className={clsx(styles.videoWrapper)} 
           />
-          {showPlayPauseButton && (
-            <button className={clsx(styles.playButton)} onClick={handlePlayPause}>
-              {isPlaying ? "Pause" : "Play"}
-            </button>
-          )}
         </>
       )}
     </MediaPlayerContextWrapper>
--- a/src/components/video-player/index.js
+++ b/src/components/video-player/index.js
@@ -6,7 +6,6 @@ import {
   useMediaStartEventTracker,
   useMediaPauseEventTracker,
   useMediaStopEventTracker,
-  useFailureEventTracker,
 } from "@objectiv/tracker-react";
 
 import clsx from "clsx";
@@ -23,7 +22,6 @@ function VideoPlayer({
   const trackMediaStart = useMediaStartEventTracker();
   const trackMediaPause = useMediaPauseEventTracker();
   const trackMediaStop = useMediaStopEventTracker();
-  const trackMediaFailure = useFailureEventTracker();
 
   function handlePlayPause() {
     setPlaying(!isPlaying);
@@ -39,7 +37,6 @@ function VideoPlayer({
         onStart={trackMediaStart}
         onPause={trackMediaPause}
         onEnded={trackMediaStop}
-        onError={trackMediaFailure}
       />
       {showPlayPauseButton && (
         <button className={clsx(styles.playButton)} onClick={handlePlayPause}>
--- a/src/components/video-player/index.js
+++ b/src/components/video-player/index.js
@@ -1,11 +1,11 @@
 import React, { useState } from "react";
 import ReactPlayer from "react-player";
-import { MediaPlayerContextWrapper } from "@objectiv/tracker-react";
 import {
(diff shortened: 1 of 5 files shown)


src/components/video-player/index.js
```diff
@@ -0,0 +1,50 @@
...
+      {showPlayPauseButton && (
+        <button className={clsx(styles.playButton)} onClick={handlePlayPause}>
+          {isPlaying ? "Pause" : "Play"}
+        </button>
+      )}
+    </MediaPlayerContextWrapper>
+  );
+}
+
+export default VideoPlayer;
```
sdirosa, line 50: @ivarpruijn 

Very nice that you stumbled into this.  

This cannot work correctly. But I can see how it may look correct without knowing how React works.

And it's interesting you stumbled into this case immediately because: 
1. It's something I've talked about in my presentation last Tuesday.
2. I was doubting whether to write an how-to about it. I think I'm sure I have to now.

---

**Why this doesn't work**
Here goes a long explanation. Feel free to skip if not interested; but I think every React developer should naturally understand and know this.

The main source of confusion is that React hides the fact that **JSX is just JavaScript executing like any other function in the body of the Component**.  

When you look at this Component it comes almost natural to think that there's 
1. a section above that calls some hooks, and defines some handlers 
2. there's a section below, that defines what the component renders, with some JSX

This actually makes no sense. JSX is very much JavaScript as well. Babel will convert all of those fake HTML tags to actual `React.createElement` calls during compile. 

Eventually the whole Component will be just one big JavaScript function calling some functions in the body and calling some more functions in the `return` statement below that compiles the JSX. 

And, of course, everything will execute exactly in the order it is. Top down. 

This means the hooks are always executed before the `React.createElement` that renders the `MediaPlayerContextWrapper` component. 

There's no way that `trackMediaLoad` knows about `MediaPlayerContextWrapper` since when that handler is produced by the hook call, `MediaPlayerContextWrapper` doesn't even exist yet.

I personally think this is the single most confusing aspect of React functional Components and hooks. I like to believe that this is the source of half of the bugs in React Applications out there 😆 

---

**The lame solution**
Split the VideoPlayer and the hooks call in a sub-Component. This will ensure that React **knows** about the Wrapper when the hooks are invoked. 

Basically because React will Render the Wrapper Component first, and then render the VideoPlayer+hooks after. The hooks will execute after the Wrapper has been processed. 

I don't like this solution (for a number of reasons, among which readability) and prefer instead to use render-props.

---

**The proper solution**
My preferred solution, and the one I'd like to suggest in our how-to-guide as well, is instead to completely scrap using the hooks (they were not meant for this case anyway) and instrument the Events with low-level Event Trackers and Wrapper render-props.


The changes are minimal:
- Import the low level Event Trackers instead of the hooks
- Use the LocationWrapper render-props to retrieve the trackingContext
- Wrap the original JSX in a fragment 
- Hook the events of VideoPlayer to the EventTrackers, passing to them the trackingContext

Here is the full code:

```typescript
import {
  MediaPlayerContextWrapper,
  trackMediaLoadEvent,
  trackMediaPauseEvent,
  trackMediaStartEvent,
  trackMediaStopEvent
} from "@objectiv/tracker-react";

import clsx from "clsx";
import React, { useState } from "react";
import ReactPlayer from "react-player";
import styles from "./styles.module.css";

function VideoPlayer({
  url,
  id = "video",
  enableControls = true,
  showPlayPauseButton = false,
}) {
  const [isPlaying, setPlaying] = useState(false);

  function handlePlayPause() {
    setPlaying(!isPlaying);
  }

  return (
    <MediaPlayerContextWrapper id={id} className={styles.videoWrapper}>
      {(trackingContext) => (
        <>
          <ReactPlayer
            url={url}
            playing={isPlaying}
            controls={enableControls}
            onReady={() => trackMediaLoadEvent(trackingContext)}
            onStart={() => trackMediaStartEvent(trackingContext)}
            onPause={() => trackMediaPauseEvent(trackingContext)}
            onEnded={() => trackMediaStopEvent(trackingContext)}
          />
          {showPlayPauseButton && (
            <button className={clsx(styles.playButton)} onClick={handlePlayPause}>
              {isPlaying ? "Pause" : "Play"}
            </button>
          )}
        </>
      )}
    </MediaPlayerContextWrapper>
  );
}

export default VideoPlayer;
```

jansentom: whoa! Awesome guide! Makes for some light reading before bedtime ;)

sdirosa: Hahaha. I spoil the fanciness of JSX and tell how lame it really is ;)
The clickbait title could be "React creators HATE 😠 this developer for what he revealed!!1".
Apart from the jokes, most of it it's pretty logical I think.

sdirosa: I'm adding a Render Props example to all of the Wrapper documentation pages. Afterwards I plan to write also a more in-depth how-to.

Screen Shot 2022-01-28 at 10 37 14

sdirosa: Docs update with lots of content around this topic: https://github.com/objectiv/objectiv.io/pull/252

Doing a round of checks and then I'll hit the merge button. I think this may be worth posting also in Announcements in Slack, as it contains useful stuff for the brave souls that are testing the ReactTracker now.

ivarpruijn:
@sdirosa Unfortunately there doesn't seem to be a way to track the CTA at the end of the video from the player SDK (or at least Vimeo support doesn't think so), so I think we can merge this PR as-is.

ivarpruijn:
Will merge this PR now, so we can get the Vimeo version to production and re-run cookiebot.

PR #255, opened 2022-01-31
Fix team links on About page

ivarpruijn:
Fixes the links on the About page for the core team.

@sdirosa Seems like the `href` option isn't working correctly for the custom `TrackedLink` component.

### DIFF
--- a/src/pages/about.tsx
+++ b/src/pages/about.tsx
@@ -19,7 +19,7 @@ function Contributor({name, gitHubUsername}) {
       <TrackedDiv id={'contributor-card'}>
         <div className="card__header">
           <TrackedDiv id={'avatar'} className={clsx("avatar", styles.contributorAvatar)}>
-            <TrackedLink href={ghProfileLink} title={ghProfileTitle}>
+            <TrackedLink to={ghProfileLink} title={ghProfileTitle}>
               <Avatar 
                 githubHandle={gitHubUsername} 
                 size='64'
@@ -32,7 +32,7 @@ function Contributor({name, gitHubUsername}) {
           </TrackedDiv>
           <div className="avatar__intro">
             <TrackedDiv id={'avatar-subtitle'} className={clsx(styles.contributorAvatarSubtitle)}>
-              <TrackedLink href={ghProfileLink} title={ghProfileTitle}>
+              <TrackedLink to={ghProfileLink} title={ghProfileTitle}>
                 @{gitHubUsername}
               </TrackedLink>
             </TrackedDiv>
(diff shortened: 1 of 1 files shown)


src/pages/about.tsx
```diff
@@ -32,7 +32,7 @@ function Contributor({name, gitHubUsername}) {
           </TrackedDiv>
           <div className="avatar__intro">
             <TrackedDiv id={'avatar-subtitle'} className={clsx(styles.contributorAvatarSubtitle)}>
-              <TrackedLink href={ghProfileLink} title={ghProfileTitle}>
+              <TrackedLink to={ghProfileLink} title={ghProfileTitle}>
```
borft, line 35: shouldn't this be a `TrackedAnchor`?  as per https://objectiv.io/docs/tracking/react/api-reference/trackedElements/TrackedAnchor

ivarpruijn: Surai made a `TrackedLink` custom component specifically for Docusaurus. See also the tip at https://objectiv.io/docs/tracking/react/how-to-guides/tracking-interactions

ivarpruijn: Actually, I think `TrackedAnchor` doesn't work. @sdirosa could you have a look?

borft: ah check. But if the code is:
```javascript
export const TrackedLink = React.forwardRef<HTMLAnchorElement, TrackedLinkProps>((props, ref) => (
  <TrackedLinkContext Component={Link} {...props} href={props.href ?? props.to} ref={ref}/>
))

```
why would this change work? Doesn't it accept `href` or `to`?

sdirosa: TrackedAnchor would work but it's not recommended by Docusaurus as you would skip all the automation around link checking. 

By using Docusaurus Links for everything, and not directly `<a>` tags, you actually get the Docusuarus link checker to check those for you, including whether they are external, etc.

TrackedLink is a 1-line custom Component that wraps around those.

The main difference between using `to` or `href` is whether you want to point to an internal location or an external one: 
- to: router
- href: anchor

The prop you use determine what gets rendered basically.

borft: ah, makes complete sense. Thanks for clarifying.

sdirosa: I double checked this case, and I think Docusaurus changed how Link behaves, so I need to update the TrackedLink component accordingly. 

In the previous version you could specify either `href` or `to` or both and the component would be smart enough to understand what it was by looking at its value, but that doesn't seem to be the case. If I specify href and pass an internal link it doesn't work as intended, and vice-versa.

I'll make a PR to fix TrackedLink up

borft: Cool, I guess we should also update the howto in that case

sdirosa: Indeed, I'm on it already

sdirosa: Example is also up to date now

sdirosa:
@ivarpruijn I think this PR is not necessary now that I fixed TrackedLink to work correctly with external links

ivarpruijn:
> @ivarpruijn I think this PR is not necessary now that I fixed TrackedLink to work correctly with external links

@sdirosa Yep, that fixes it. Closing this PR. Thanks.

PR #262, opened 2022-02-01
Update Docusaurus TrackedLink example with latest code

sdirosa:
This PR updates the how-to guide mentioning Docusaurus TrackedLink Component to the latest version of the source code.

### DIFF
--- a/docs/docs/tracking/react/how-to-guides/tracking-interactions.md
+++ b/docs/docs/tracking/react/how-to-guides/tracking-interactions.md
@@ -90,7 +90,12 @@ export type TrackedLinkProps = Omit<TrackedLinkContextProps, 'Component' | 'href
 
 export const TrackedLink = React.forwardRef<HTMLAnchorElement, TrackedLinkProps>(
   (props, ref) => (
-    <TrackedLinkContext Component={Link} {...props} href={props.href ?? props.to} ref={ref}/>
+    <TrackedLinkContext 
+      Component={Link} {...props} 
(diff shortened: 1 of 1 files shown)


docs/docs/tracking/react/how-to-guides/tracking-interactions.md
```diff
@@ -90,7 +90,12 @@ export type TrackedLinkProps = Omit<TrackedLinkContextProps, 'Component' | 'href
 
 export const TrackedLink = React.forwardRef<HTMLAnchorElement, TrackedLinkProps>(
   (props, ref) => (
-    <TrackedLinkContext Component={Link} {...props} href={props.href ?? props.to} ref={ref}/>
+    <TrackedLinkContext 
+      Component={Link} {...props} 
+      href={props.href ?? props.to} 
+      forwardHref={!!props.href} 
+      ref={ref}
+    />
```
sdirosa, line 98: Only difference here is that Link now requires the `href` attribute to be values for external links, while before it was possible to pass that as `to` as well. 

I guess it was not reliable enough to try and guess what a link is. I actually like that this is explicit now. Or maybe it was never intended behavior and it got fixed.

For reference, `forwardHref` is a `TrackedLinkContext` prop that allows developers to forward `href` to the underlying component. 

There is also a `forwardId` prop for the same reason.

Since we cannot determine automatically if `id` or `href` where intended to be passes to the given `Component` or not, we don't do it by default and we allow to override that behavior via props.

PR #265, opened 2022-02-01
Fix issue building on Windows

ivarpruijn:
If an `import` in an `md` file is done before the title, the sidebar menu doesn't render the correct title, when making a build on Windows. This works fine on a Linux system.

### DIFF
--- a/docs/docs/tracking/react/how-to-guides/custom-components.md
+++ b/docs/docs/tracking/react/how-to-guides/custom-components.md
@@ -2,9 +2,9 @@
 sidebar_position: 5
 ---
 
-import useBaseUrl from '@docusaurus/useBaseUrl';
-
 # Custom Components
+
+import useBaseUrl from '@docusaurus/useBaseUrl';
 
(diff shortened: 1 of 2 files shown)


docs/docs/tracking/react/how-to-guides/custom-components.md
```diff
@@ -2,10 +2,10 @@
 sidebar_position: 5
 ---
 
-import useBaseUrl from '@docusaurus/useBaseUrl';
-
 # Custom Components
 
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
```
sdirosa, line 8: I think this is not even used. But what a weird bug! 😮

sdirosa: We were guessing with Mike. 

My theory is that there's some logic to ignore the import to get the first line of the MD but it doesn't take in account Windows CR+LF. So the first line is actually some empty line and Docusaurus reverts to printing out the slug.

ivarpruijn: Seems likely. Still bad. I'll merge this PR now so it 'fixes' it.

PR #268, opened 2022-02-01
Vimeo player

ivarpruijn:


src/components/vimeo-player/index.js
```diff
@@ -0,0 +1,35 @@
...
+          app_id='58479'
+          onReady={() => trackMediaLoadEvent(trackingContext)}
+          onPlay={() => trackMediaStartEvent(trackingContext)}
+          onPause={() => trackMediaPauseEvent(trackingContext)}
+          onEnd={() => trackMediaStopEvent(trackingContext)}
+          className={clsx(styles.videoWrapper)}
+          />
+        </>
+      )}
+    </MediaPlayerContextWrapper>
```
sdirosa, line 31: Tested this on dev both in incognito and normal mode, works like a charm. No cookies are set and the video is totally usable.

PR #269, opened 2022-02-01
Use locally imported Vimeo player for SSR compatibility

sdirosa:
@ivarpruijn

### DIFF
--- a/src/components/vimeo-player/index.js
+++ b/src/components/vimeo-player/index.js
@@ -1,5 +1,5 @@
-import BrowserOnly from "@docusaurus/core/lib/client/exports/BrowserOnly";
-import React, { useState } from "react";
+import BrowserOnly from '@docusaurus/BrowserOnly';
+import React from "react";
 import clsx from "clsx";
 import styles from "./styles.module.css";
 import {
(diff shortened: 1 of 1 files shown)


src/components/vimeo-player/index.js
```diff
@@ -9,25 +10,26 @@ import {
...
-          onPlay={() => trackMediaStartEvent(trackingContext)}
-          onPause={() => trackMediaPauseEvent(trackingContext)}
-          onEnd={() => trackMediaStopEvent(trackingContext)}
-          className={clsx(styles.videoWrapper)}
-          data-cookieconsent="ignore"
-          />
-        </>
+        <BrowserOnly>
+          {() => {
+            const Vimeo = require('@u-wave/react-vimeo').default;
```
sdirosa, line 19: This library doesn't export Vimeo as a component but as `default`

src/components/vimeo-player/index.js
```diff
@@ -1,3 +1,4 @@
+import BrowserOnly from "@docusaurus/core/lib/client/exports/BrowserOnly";
```
sdirosa, line 1: @ivarpruijn PyCharm doesn't understand docusaurus at all, this import may be shortened I think

sdirosa: Fixed

PR #271, opened 2022-02-02
build website + docs in docker container and deploy the result

borft:
We use de node:16 container as the build environment to produce stable and reproducable builds. Additionally, the `objectiv/website` container now uses these build artefacts as well.

Use `--build-arg` to specify `OBJECTIV_ENVIRONMENT` to determine the environment to build for, when using docker directly. When using the Makefile, simple use the environment variable. The default is to build for docker.

How to build the website
```bash
# build image. By default this builds for staging and production
make build-docker-build-image

# manual build (make sure to build docker/build/Dockerfile first):
docker build --no-cache -t objectiv/website-build -f docker/build/Dockerfile .

# build website image, this will first build for docker, and then build the
# website container using that result.
make  build-docker-website-image

# manual build (make sure to build docker/build/Dockerfile first):
docker build --no-cache -t objectiv/website -f docker/website/Dockerfile docker
```

Additionally, this adds a GH action, that is called on merges to main. This triggers building for both production and staging. Both will be uploaded to a temporary folder. Staging will be activated. To update production, some manual intervention is still needed.
To facilitate deployment, there's a third image (objectiv/website-deploy) that incorporates `lftp` which is used to push the build artefacts to the hosting provider.

FYI: Running a full build + deployment locally (using the github action) takes about 10 minutes.


Makefile
```diff
@@ -1,42 +1,34 @@
...
 
 # build docker container for full website, including docs
 # set environment to docker, to make sure the right config/env is loaded
-build-docker-website: build/index.html docs/build/index.html
-	docker build --no-cache -t objectiv/website -f docker/Dockerfile .
+build-docker-website-image: OBJECTIV_ENVIRONMENT = "docker"
+build-docker-website-image: build-docker-build-image
+	docker build --no-cache -t objectiv/website -f docker/website/Dockerfile docker
 
+build-docker-deploy-image:
```
borft, line 26: there's no hard dependency for the build-docker-build-image here, so it will deploy the most recent build there is.

docker/deploy/deploy.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [ -n "${SFTP_PUBKEY}" ];
+then
+  echo "Writing pubkey to known_hosts"
+  mkdir -p ~/.ssh
+  echo "${SFTP_PUBKEY}" >> ~/.ssh/known_hosts
+fi
+#
+SCRIPT="
+lcd /services
```
borft, line 14: this is where the deployment magic happens

Makefile
```diff
@@ -1,42 +1,34 @@
...
+	docker build --no-cache -t objectiv/website-deploy -f docker/deploy/Dockerfile .
 
 # spin up the website container, and check all _internal_ links for broken ones
 # external links are skipped
-check-broken-links: build-docker-website
+check-broken-links: build-docker-website-image
 	# spin up website
 	docker run --rm -d -p 127.0.0.1:8080:80 --name objectiv_website_broken_link_check objectiv/website
 	./node_modules/.bin/blc --recursive --exclude-external --ordered --host-requests 10 http://localhost:8080
```
sdirosa, line 34: Just an idea, but wouldn't it be nice to run the link checker automatically to make sure we never upload builds with broken links? 

Perhaps before uploading anything to staging?

sdirosa:
@borft This looks surprisingly easy to follow. Nice work!
Do you think it makes sense for me to test this manually or perhaps would it be best to just merge this up and use it? 

Since It only pushes to staging, if it needs some polishing there's no real problem, right?

I've left a minor suggestion, nothing blocking though.

Makefile
borft: good point, let's add it

borft:
> @borft This looks surprisingly easy to follow. Nice work! Do you think it makes sense for me to test this manually or perhaps would it be best to just merge this up and use it?
> 
> Since It only pushes to staging, if it needs some polishing there's no real problem, right?
> 
> I've left a minor suggestion, nothing blocking though.

Yes, the tricky part is, that staging and production deployments share the same hosting. But I've tested it quite a bit on a different domain, and as far as I can tell, it does what it is supposed to.

Makefile
borft: hmm, this is kind of hard for a number of reasons:
- to run the check, the website needs to be served, so we need an image with an httpd
- running dockerized means we need yet another build
- because connections are internal, there needs to be an additional config, specifically for this environment

borft: What I've done now, is add the broken link checker after deploying to staging. That way, hosting etc is already taken care off, and we'd still be notified of broken links. Not the best, but better than nothing IMHO.

borft: yeah, serve doesn't behave the same, also, it doesn't work, because we need both docs and website simultaneously, and finally, the website container doesn't have `node`

PR #273, opened 2022-02-02
Upgrade blog components to Docusaurus beta15

ivarpruijn:
- Upgrade blog components to Docusaurus beta15. Only some internal refactoring by Docusaurus, nothing that was broken.
- Remove `theme-classic` dependency, as it's already imported by the `preset-classic` dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
   "dependencies": {
     "@docusaurus/core": "^2.0.0-beta.15",
     "@docusaurus/preset-classic": "^2.0.0-beta.15",
-    "@docusaurus/theme-classic": "^2.0.0-beta.15",
     "@mdx-js/react": "^1.6.22",
     "@objectiv/tracker-react": "^0.0.7",
     "@u-wave/react-vimeo": "^0.9.5",
--- a/src/theme/BlogPostItem/index.tsx
+++ b/src/theme/BlogPostItem/index.tsx
@@ -12,6 +12,7 @@ import {MDXProvider} from '@mdx-js/react';
 import Translate, {translate} from '@docusaurus/Translate';
 import {useBaseUrlUtils} from '@docusaurus/useBaseUrl';
 import {usePluralForm} from '@docusaurus/theme-common';
+import {blogPostContainerID} from '@docusaurus/utils-common';
 import MDXComponents from '@theme/MDXComponents';
 import EditThisPage from '@theme/EditThisPage';
 import type {Props} from '@theme/BlogPostItem';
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -17,7 +17,6 @@
   "dependencies": {
     "@docusaurus/core": "^2.0.0-beta.15",
     "@docusaurus/preset-classic": "^2.0.0-beta.15",
-    "@docusaurus/theme-classic": "^2.0.0-beta.15",
```
sdirosa, line 20: Nice 👌🏼

sdirosa:
@ivarpruijn There are other dependencies that are fairly outdated. Do we want to bump those up in this PR or perhaps in a clean one?

ivarpruijn:
> @ivarpruijn There are other dependencies that are fairly outdated. Do we want to bump those up in this PR or perhaps in a clean one?

Good one, but think we should do that in another PR then.

src/theme/BlogPostItem/index.tsx
```diff
@@ -77,17 +78,25 @@ function BlogPostItem(props: Props): JSX.Element {
...
+      itemScope
+      itemType="http://schema.org/BlogPosting">
       <TrackedHeader>
         <TitleHeading className={styles.blogPostTitle} itemProp="headline">
           {isBlogPostPage ? (
             title
           ) : (
-            <TrackedLink itemProp="url" to={permalink}>
+            <TrackedLink 
+              id={`post-${makeIdFromString(title)}`}
```
sdirosa, line 97: This identifier should not be unique, the ContentContext around this whole Component already describes in which post we are and the TrackedHeader ensures other links, identical to this one, will not collide with it.

src/theme/PaginatorNavLink/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+function PaginatorNavLink(props: Props): JSX.Element {
+  const {permalink, title, subLabel} = props;
+  return (
+    <TrackedLink className="pagination-nav__link" to={permalink}>
+      {subLabel && <div className="pagination-nav__sublabel">{subLabel}</div>}
+      <div className="pagination-nav__label">{title}</div>
+    </TrackedLink>
+  );
+}
```
sdirosa, line 21: Nice Component!

src/theme/BlogPostItem/index.tsx
ivarpruijn: You're right. I was testing with a 2nd blog post with the exact same title, but of course that collides. That will never happen in practice, so I'll remove this ID.

sdirosa: Indeed, that's the only scenario but it's really a strange case imho, the routing wouldn't work either and the second article would never be reachable

sdirosa (approved):

ivarpruijn:
Tested on staging, ran broken links checker, all lgtm

Issue #277, opened 2022-02-03
Mermaid graphs aren't rendering on website


smoens:
**Describe the bug**
Mermaid graphs are shown in plain text instead of visualized.
e.g. URL [Objective - Global Contexts](https://objectiv.io/docs/taxonomy/global-contexts/)

**Expected behavior**
Mermaid graph is expected to be rendered

**Screenshots**
![image](https://user-images.githubusercontent.com/1371244/152298052-4fdf7e94-a976-4e44-810f-6f20497c8b2f.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox 96.0.3

ivarpruijn:
@smoens Thanks for reporting this! We're looking into it. Seems it's related to an overly eager cookie consent that's blocking Mermaid from loading.

ivarpruijn:
@smoens A fix is live now. Thanks again for reporting, and let us know if you run into any issues!

PR #278, opened 2022-02-03
Fix Mermaid graphs not rendering without cookie consent

ivarpruijn:
Reported in https://github.com/objectiv/objectiv.io/issues/277. Cookiebot is blocking Mermaid graphs from loading, even though it doesn't set cookies. 

Issue can be reproduced on production, but not in dev mode, nor on staging.

This potential (since we can't reproduce locally) fix:
- Sets Cookiebot's blocking mode to manual. We don't use anything that sets cookies other than our own tracking, so this should be fine.
- Adds `data-cookieconsent="ignore"` to the element containing the graph, although most likely that doesn't do anything.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -62,7 +62,7 @@ const config = {
     {
       src: 'https://consent.cookiebot.com/uc.js?cbid=7498452c-872b-431a-9859-21045f83f0a0',
       'data-cbid': '7498452c-872b-431a-9859-21045f83f0a0',
-      'data-blockingmode': 'auto',
+      'data-blockingmode': 'manual',
       id: 'Cookiebot'
     },
   ],
--- a/docs/src/theme/Mermaid/index.tsx
+++ b/docs/src/theme/Mermaid/index.tsx
@@ -13,6 +13,7 @@ mermaid.initialize({
     curve:'cardinal',
     useMaxWidth: true,
   },
+  dataCookieConsent: 'ignore',
   theme: 'base', 
   themeCSS: `
     .node rect { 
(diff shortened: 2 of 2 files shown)


docs/src/theme/Mermaid/index.tsx
```diff
@@ -13,6 +13,7 @@ mermaid.initialize({
     curve:'cardinal',
     useMaxWidth: true,
   },
+  dataCookieConsent: 'ignore',
```
ivarpruijn, line 16: Mermaid documentation is broken atm, so not sure this is used.

docs/src/theme/Mermaid/index.tsx
```diff
@@ -62,7 +63,7 @@ const Mermaid = ({ chart, caption, baseColor, links}) => {
 
 	return (
     <div className={"diagram-"+baseColor}>
-      <div className="mermaid">{chart}</div>
+      <div className="mermaid" data-cookieconsent="ignore">{chart}</div>
```
ivarpruijn, line 66: Likely doesn't do anything, but just to be sure.

docs/docusaurus.config.js
```diff
@@ -62,7 +62,7 @@ const config = {
     {
       src: 'https://consent.cookiebot.com/uc.js?cbid=7498452c-872b-431a-9859-21045f83f0a0',
       'data-cbid': '7498452c-872b-431a-9859-21045f83f0a0',
-      'data-blockingmode': 'auto',
+      'data-blockingmode': 'manual',
```
ivarpruijn, line 65: Should be fine, as we don't use any external scripts that set cookies.

ivarpruijn:
Merging this into #275 to test all changes

PR #279, opened 2022-02-03
Set cookiebot blocking mode to manual

ivarpruijn:
To align with the docs, set Cookiebot's blocking-mode to manual. 

Any script that relies on the window 'load' event will just run instead of getting blocked, unless we control it manually. This is fine for the website, which doesn't rely on external libraries other than Vimeo, which is currently manually set to always run in DNT mode.


sdirosa (approved):
This seems to work nicely 👌🏼

ivarpruijn:
Tested on staging, looks good, merging

PR #281, opened 2022-02-03
Fix docusaurus links from modeling to outside modeling

borft:
- add trailing slash to slug for introduction/index
- make sure to only insert "modeling" in a.href if window .location ends with modeling without a trailing slash

### DIFF
--- a/docs/docs/modeling/intro.mdx
+++ b/docs/docs/modeling/intro.mdx
@@ -2,7 +2,7 @@
 id: intro
 title: Introduction — Bach  documentation
 hide_title: true
-slug: /modeling
+slug: /modeling/
 sidebar_label: Introduction
 
 ---
--- a/docs/src/components/sphinx-page/index.tsx
+++ b/docs/src/components/sphinx-page/index.tsx
@@ -124,7 +124,8 @@ const SphinxPage = (props) => {
                 });
 
                 // get base from window.location, should be something like https://objectiv.io, or http://localhost:3000
-                const currentSite = window.location.toString().match(/^(http[s]?:\/\/[a-z0-9.:]+\/).*?$/);
+                const currentLocation = window.location.toString();
+                const currentSite = currentLocation.match(/^(http[s]?:\/\/[a-z0-9.:]+\/).*?$/);
 
                 // fix anchors (remove .html) and fix path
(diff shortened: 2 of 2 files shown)


docs/docs/modeling/intro.mdx
```diff
@@ -2,7 +2,7 @@
 id: intro
 title: Introduction — Bach  documentation
 hide_title: true
-slug: /modeling
+slug: /modeling/
```
borft, line 5: this makes sure we keep the trailing slash when clicking Introduction.  see: https://github.com/objectiv/objectiv-analytics/pull/437/files for the corresponding fix

docs/src/components/sphinx-page/index.tsx
```diff
@@ -136,9 +137,14 @@ const SphinxPage = (props) => {
 
                     if ( isInternal ){
                         // fix the hrefs in the overview/index page in case of missing trailing slash
-                        if ( a.href.indexOf('modeling') == -1 && window.location.toString().slice(-1) != '/' ){
+                        // but only if we're toplevel, eg.
+                        // https://$currentSite/.../modeling and there's no modeling in a.href
+                        if ( a.href.indexOf('modeling') == -1
+                            && currentLocation.endsWith('modeling')
```
borft, line 143: extra condition here, we only do this, if the current location ends in `modeling`, so without the trailing slash.

ivarpruijn:
LGTM, also on staging

docs/src/components/sphinx-page/index.tsx
```diff
@@ -123,22 +123,32 @@ const SphinxPage = (props) => {
...
 
                 // fix anchors (remove .html) and fix path
                 Object.values(tempDiv.getElementsByTagName('a')).forEach( a => {
 
                     // a link is internal if the first part matches the current location,
+                    // or the tld's match (staging vs production
                     // or if it's a relative URL
-                    const isInternal = ((currentSite[1] !== undefined && a.href.startsWith(currentSite[1])) ||
+                    const isInternal = ( (currentSite[1] !== undefined && a.href.startsWith(currentSite[1])) ||
+                        currentDomain == extractDomain(a.href) ||
```
borft, line 141: if the tld's match, links are now also considered internal. Which fixes the objectiv.io vs staging.objectiv.io conumdrum

PR #284, opened 2022-02-03
Update getting-started.md

gertjanspriensma:
I would use collector-url or something like that, by using /collector it looks like you have to append the localhost:5000 with /collector.

### DIFF
--- a/docs/docs/tracking/react/how-to-guides/getting-started.md
+++ b/docs/docs/tracking/react/how-to-guides/getting-started.md
@@ -28,7 +28,7 @@ import { ObjectivProvider, ReactTracker } from '@objectiv/tracker-react';
 const App = ({children}) => {
 
   const tracker = new ReactTracker({
-    endpoint: '/collector',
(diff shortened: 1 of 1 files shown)


ivarpruijn:
Thanks @gertjanspriensma! LGTM

sdirosa (approved):
Looks good. 

For context. This example was based on a CRA setup where we used its proxy option to map /collector to the actual collector BE via .env.

This allows for a single artifact build that can be used either for dev, staging or prod without needing to rebuild.

ivarpruijn:
This is live now, thanks @gertjanspriensma : https://objectiv.io/docs/tracking/react/how-to-guides/getting-started

sdirosa:
Follow up on this: https://github.com/objectiv/objectiv.io/pull/286

The confusing `endpoint` prop was all over the place. I have replaced it with an actual consistent URL everywhere. I think this should rule out pretty clearly what it actually is. 

I plan to explain properties one by one in a future documentation update. Including examples.

PR #287, opened 2022-02-07
Docs: Use second slug as RootLocationContext ID + 'home' for first docs section

sdirosa:
## Changes
- Reconfigured the BrowserTracker RootLocationContextFromURLPlugin of the Docs to determine the identifier of RootLocationContext from the second slug of the url instead of the first, as the latter is always `docs`. This will produce correct RootLocationContexts for the Docs.
- Updated slugs of main docs section to `/home/etc.`, so they always produce 'home' as a RootLocationContext ID.
- Updated all absolute links to relative file links, so Docusaurus can detect any broken links in the future.
- Updated website links to the affected docs pages.

**Extras**. 
- Reconfigured dev env to mount as /docs, to make it possible to test these changes

## TODO
- [ ] Deploy docs & website together, so links don't break.


ivarpruijn (approved):
@sdirosa Added my changes. Let's test site & docs in one go on staging, and make sure we deploy them together as well. FYI @borft

sdirosa:
> @sdirosa Added my changes. Let's test site & docs in one go on staging, and make sure we deploy them together as well. FYI @borft

I'm going to double check each link asap

docs/docs/quickstart-guide.md
```diff
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 2
-slug: /quickstart-guide
+slug: /home/quickstart-guide
```
sdirosa, line 3: This makes it harder to maintain I think, because the link checker of Docusaurus doesn't know about frontmatter.

For example this link is present in several places and currently broken:

1. - docs/introduction.md
2. - FAQ.md
3. - tracking/collector/getting-started.md
4. - examples page (demo link) of the statically generated modeling docs
5. - intro page (demo link) for the modeling docs

I think it would be easier to move all of these pages in a `home` directory allowing Docusuarus to detect, at least, the first 4 pages

sdirosa, line 3: This makes it harder to maintain I think, because it looks like the link checker of Docusaurus doesn't know about frontmatter.

For example this link is present in several places and currently broken:

1. - docs/introduction.md
2. - FAQ.md
3. - tracking/collector/getting-started.md
4. - examples page (demo link) of the statically generated modeling docs
5. - intro page (demo link) for the modeling docs

I think it would be easier to move all of these pages in a `home` directory allowing Docusuarus to detect, at least, the first 4 pages

ivarpruijn: Right, this is why Docusaurus recommends to use relative (file) links to docs, which I guess we didn't nicely do everywhere: https://docusaurus.io/docs/docs-markdown-features#referencing-other-documents

I'm fine with moving everything into a separate directory, though it means we'll have to update links in several places anyway. Note taht the `/introduction.md` file probably can't be moved, as there's at least one root page needed.

borft:
There's also a few links in the modeling section to the quickstart, which are broken now I think.

docs/docs/quickstart-guide.md
sdirosa: The Tracking section only uses relative file links everywhere afaik. If not that's a mistake and should be fixed.

Is that not the case for the other sections?

Regarding the introductory page, that's fine I think; because it doesn't have a slug of its own. Hopefully we haven't linked to it directly.

Do you need help fixing all these links?

ivarpruijn: I'll fix it up, will probably be tomorrow

sdirosa: Ok, I've double checked the /tracking section and found a few links that did not have filenames in their urls. Fixed all of them except one; case #3 above.

docs/.env.development
```diff
@@ -1,5 +1,5 @@
 trackerApplicationId="objectiv-docs-dev"
 trackerEndPoint="http://localhost:5000"
 websiteUrl="http://localhost:3000"
-baseUrl="/"
+baseUrl="/docs/"
```
borft, line 4: this is nice, makes it more in line with the other environments

sdirosa: It was also the only way I could test these URLs while in development. Docusaurus is actually quite smart and boots the docs as /docs automatically like this, so you can check the LocationStacks as they would be on staging / production

docs/src/theme/Root.tsx
```diff
@@ -66,7 +66,7 @@ function Root({children}) {
 
       const trackerPlugins = new TrackerPlugins({
         plugins: [
-          ...makeTrackerDefaultPluginsList(trackerOptions),
+          ...makeDefaultPluginsList(trackerOptions),
```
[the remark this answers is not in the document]
ivarpruijn: OK, makes sense then. Thanks!

docs/src/theme/Root.tsx
```diff
@@ -64,7 +67,20 @@ function Root({children}) {
...
+            new PathContextFromURLPlugin(trackerOptions),
+            new RootLocationContextFromURLPlugin({
+              ...trackerOptions,
+              idFactoryFunction: () => {
+                const secondSlug = location.pathname.split('/')[2];
+
+                return secondSlug ? secondSlug.trim().toLowerCase() : 'home';
+              }
+            })
+          ]
```
sdirosa, line 83: @ivarpruijn I plan to change the new configuration syntax again. To avoid having to redeploy again when that happens, I reverted this to the low level way of configuring plugins. This way when I change the API again it will not affect the docs.

ivarpruijn:
@sdirosa Pushed both the site + docs to staging from this branch, and tested it. All but one thing checks out: the links to the quickstart guide in the Modeling section need to be updated. Did that in this PR: https://github.com/objectiv/objectiv-analytics/pull/471/files. Have asked @borft to rerun the docs and commit them in this branch, so it can all go live in one go.

ivarpruijn:
Tested on staging, looks good, merging

PR #290, opened 2022-02-10
Add 'Objectiv in plain English' video to homepage

ivarpruijn:


jansenbob:
apart from the discussed black bars, LGTM. Not a blocker in my book

PR #295, opened 2022-02-14
updating HttpContext for new referrer naming

vincenthoogsteder:


docs/docs/taxonomy/reference/global-contexts/HttpContext.md
```diff
@@ -22,7 +22,7 @@ import Mermaid from '@theme/Mermaid';
 | :--       | :--         | :--           
 | **id**    | string      | Unique string to be combined with the Context Type (`_type`) for Context instance uniqueness.
 | **_type** | string      | String literal used during serialization. Should always match the Context interface name.          
-| **referer**         | string          | Full URL to HTTP referrer of the current page.
+| **referrer**         | string          | Previous URL that the user visited before the event.
```
borft, line 25: So technically, the value in here, is what the browser sets in the referer header in the request that loaded the current page. Where current page, is the page that generated the event. Usually that will be the URL to the previous page. 

But I think it's important to stress it is based on the header set by the browser. If the browser doesn't set it (for example if the link is opened in a new tab due to target=_blank), this value will be empty.

ivarpruijn:
@vincenthoogsteder ready to merge? I'll prepare a build on staging then

vincenthoogsteder:
@ivarpruijn no, awaiting final review from @borft first before we merge

borft (approved):
lgtm

PR #312, opened 2022-02-17
Use base URL for all docs links

ivarpruijn:
Use absolute URLs for all /docs links on the site, so they can also be automatically checked with the `broken-link-checker`. Also changes a few to open in the same tab, to prevent any potentials issues with iOS Safari.

Testing done:
- On staging, all links still work on both desktop and mobile (tested on Chrome only).
- Tracking events are still sent.
- Deliberately breaking all 7 adjusted links are now all actually detected by the `broken-link-checker`.
image
(8 instead of 7 because one is repeated in two blog posts)
image
image
image
image

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -137,7 +137,7 @@ export default function Home() {
               <div className={clsx(styles.outro)}>
                 <p>To learn more about the open analytics taxonomy, check out the Docs.</p>
                 <TrackedLink
-                  to={url + "/docs/taxonomy/"}
+                  to={useBaseUrl("/docs/taxonomy/", {absolute: true})}
                   autoAddBaseUrl={true}
                   waitUntilTracked={true}
                   target="_self"
@@ -248,7 +248,7 @@ export default function Home() {
               <div className={clsx(styles.outro)}>
                 <p>For an overview of all available pre-built models, check out the Docs.</p>
                 <TrackedLink
-                  to={url + "/docs/modeling/Objectiv/bach_open_taxonomy.ModelHub"}
+                  to={useBaseUrl("/docs/modeling/Objectiv/bach_open_taxonomy.ModelHub", {absolute: true})}
                   autoAddBaseUrl={true}
                   waitUntilTracked={true}
                   target="_self"
@@ -275,7 +275,7 @@ export default function Home() {
               <p className={clsx(styles.quickStartOutro)}>Follow the <strong>Quickstart Guide</strong> to 
               locally run the full Objectiv pipeline dockerized.</p>
               <TrackedLink
-                to={url + "/docs/home/quickstart-guide/"}
+                to={useBaseUrl("/docs/home/quickstart-guide/", {absolute: true})}
                 autoAddBaseUrl={true}
                 waitUntilTracked={true}
                 target="_self"
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -139,7 +139,6 @@ export default function Home() {
                 <p>To learn more about the open analytics taxonomy, check out the Docs.</p>
                 <TrackedLink
                   to={useBaseUrl("/docs/taxonomy/", {absolute: true})}
-                  autoAddBaseUrl={true}
                   waitUntilTracked={true}
                   target="_self"
                   className={clsx("button", styles.ctaButton)}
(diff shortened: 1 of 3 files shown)


src/pages/index.tsx
```diff
@@ -138,7 +138,7 @@ export default function Home() {
               <div className={clsx(styles.outro)}>
                 <p>To learn more about the open analytics taxonomy, check out the Docs.</p>
                 <TrackedLink
-                  to={url + "/docs/taxonomy/"}
+                  to={useBaseUrl("/docs/taxonomy/", {absolute: true})}
                   autoAddBaseUrl={true}
```
ivarpruijn, line 142: @sdirosa Each of the links now has the `autoAddBaseUrl` parameter; I'm not sure about its workings, is this still necessary if we use the `useBaseUrl` functionality from Docusaurus like we do here?

src/theme/BlogPostItem/index.tsx
```diff
@@ -136,12 +136,21 @@ function BlogPostItem(props: Props): JSX.Element {
...
           <p>
-            <TrackedLink to="https://objectiv.io/docs/home/quickstart-guide/">Quickstart Guide</TrackedLink> - Try 
-              Objectiv on your local machine (takes 5 minutes)<br />
-            <TrackedLink to="https://github.com/objectiv/objectiv-analytics">Objectiv on Github</TrackedLink> 
-              &nbsp;- Check out the project and star us for future reference<br />
-            <TrackedLink to={slackJoinLink as string}>Objectiv on Slack</TrackedLink> - Join the discussion 
-              or get help
+            <TrackedLink 
+              to={withBaseUrl("/docs/home/quickstart-guide/", {absolute: true})}
+              target="_self">
```
ivarpruijn, line 141: Opening all links in same tab to prevent any Safari iOS issues.

src/pages/index.tsx
sdirosa: AFAIK The `autoAddBaseUrl` option is the same as setting `to={useBaseUrl("relative-url")}`. It avoids having to import the useBaseUrl helper.

src/theme/BlogPostItem/index.tsx
sdirosa: What is the issue?

src/pages/index.tsx
sdirosa: Checked, it's exactly the same thing:
https://github.com/facebook/docusaurus/blob/main/packages/docusaurus/src/client/exports/Link.tsx#L64-L74

ivarpruijn: OK. Does it also do exactly the same as `useBaseUrl` with the `{absolute: true}` parameter? If so, we should remove one of them. I kind of like the explicit syntax then (`to={useBaseUrl("/docs/taxonomy/", {absolute: true})}`), wdyt?

src/theme/BlogPostItem/index.tsx
sdirosa: Interesting, I never experienced this issue. As a matter of fact I only have Safari on my phone and always check the website with that.

I wonder what version of iOS and Safari it was and if it's related to the latest privacy stuff they released. Would be interesting to reproduce and setup a test for these cases, so we don't have them occur anymore.

src/pages/index.tsx
sdirosa: I think it doesn't, so we should definitely remove `autoAddBaseUrl`

src/theme/BlogPostItem/index.tsx
ivarpruijn: I think it's a very recent version of both. The only links that were affected were the external links to GitHub etc. btw.

sdirosa: Let's make sure this was not a tracker induced problem and really just a misconfiuguration. I'll do some more testing by checking out the commit before the fixes and trying to reproduce on latest iOS.

ivarpruijn: @sdirosa Did you have time to reproduce on latest iOS? Are you okay with merging this, or do you want to wait?

sdirosa: Yes, the reason why target blank links don't open in Safari it's because they are automatically set as noopener. So it makes sense that those links cannot open in new windows/tabs.
So no issues with the Tracker :)

ivarpruijn: Cool, then I'll merge this PR!

PR #313, opened 2022-02-21
Troubleshooting page for Collector docs

ivarpruijn:
- Created a new Troubleshooting page for the Collector docs, inline with the other Tracking sections.
- Moved a few Troubleshooting items from the Getting Started page to this new page, and added a reference. 
- Added a new section for 'I get a FetchTransport error in the browser console'.


docs/docs/tracking/collector/troubleshooting.md
```diff
@@ -0,0 +1,75 @@
...
+  ```
+1. If the logs don't yield any clues, it's possible that the cannot listen on the given port (by default 
+port `5000`). Let's see if this is the case below.
+
+<Tabs groupId="operating-systems">
+  <TabItem value="linux" label="Linux">
+
+```bash
+TODO
+```
```
ivarpruijn, line 31: @borft could you help with this one?

docs/docs/tracking/collector/troubleshooting.md
```diff
@@ -0,0 +1,75 @@
...
+
+  </TabItem>
+  <TabItem value="mac" label="MacOS">
+
+```bash
+TODO
```
ivarpruijn, line 37: @borft could you help with this one?

docs/docs/tracking/collector/troubleshooting.md
borft: something like:
```
sudo lsof -n -i :5000
```
should give you the process (if any) that's listening on port 5000. Keep in mind, on linux the collector will refuse to start if it cannot bind to the port. On Windows/Mac this is different, because the container is running in a VM.

docs/docs/tracking/collector/troubleshooting.md
borft: probably something like:
```
netstat -an |grep 5000
```
I think @sdirosa can probably help here

sdirosa: What Mike said 👆

Getting the collector to run on macOS varies depending on the OS version though.

sdirosa: You can test this stuff in a clean environment on our BrowserStack btw, it's a bit more realistic than doing manual tests on our development machines

docs/docs/tracking/collector/troubleshooting.md
```diff
@@ -0,0 +1,75 @@
...
+At the first start-up, Postgres will be initialised. This means a database will be created. As this is 
+persisted on disk (in a docker volume), on subsequent startups, the persisted database will be loaded. In 
+case of database changes, this may cause problems (as they are also persisted). 
+
+To reset the database, simply do the following:
+```bash
+# stop the containers
+docker-compose -f docker-compose-dev.yaml down
+# remove the volume
+docker volume rm objectiv-analytics_pgdata
```
borft, line 69: actually, the name of the docker volume is determined by the directory from which the instance was started. so if your checkout dir is in fact `objectiv-analytics`, this is correct, otherwise it probably isn't.

ivarpruijn: Good addition, I added a note about that fact.

docs/docs/tracking/collector/troubleshooting.md
ivarpruijn: Thanks, added it. Also added a note about your comment.

docs/docs/tracking/collector/troubleshooting.md
ivarpruijn: Thanks, added it. Also added a note about your comment on the MacOS versions.

sdirosa (approved):
Looks good to me.

borft (commented):
collector port will move to `8081`, so might as well update that here now

borft (approved):

PR #320, opened 2022-03-03
Custom top level navigation

ivarpruijn:
We have a top level navigation in the docs with sections for Tracking, Modeling, and Taxonomy (+Home). On mobile, in the sidebar menu, those top level navigation items are currently behind a click on `Back to main menu`, which is not very user-friendly.

This PR adds a custom top-level navigation on mobile and desktop.

**Desktop:**
image

**Mobile:**
image


docs/src/theme/Navbar/objectiv.styles.module.css
```diff
@@ -0,0 +1,51 @@
+.objectivNavCategories {
+  color: #fff;
+  padding: 6px 16px;
+}
+
+.objectivNavHomepageLink {
+  margin-top: 40px;
+}
+
+/* Option A: Dropdown + title */
```
ivarpruijn, line 10: Need to choose one of these options

sdirosa:
@ivarpruijn Shall I review this PR, or better wait until it's ready?

ivarpruijn:
> @ivarpruijn Shall I review this PR, or better wait until it's ready?

Maybe a good idea to wait, I'll ping you when we've made a choice, thanks.

docs/.env.development
```diff
@@ -1,5 +1,5 @@
 trackerApplicationId="objectiv-docs-dev"
-trackerEndPoint="http://localhost:5000"
+trackerEndPoint="http://localhost:8081"
```
ivarpruijn, line 2: Update the collector to the new port on dev

docs/src/theme/NavbarItem/NavbarNavLink.tsx
```diff
@@ -36,6 +36,8 @@ export default function NavbarNavLink({
   const isDropdownLink = activeClassName === dropdownLinkActiveClass;
 
   const linkTo = href ? (prependBaseUrlToHref ? normalizedHref : href) : toUrl;
+  // @ts-ignore
+  delete props.waitUntilTracked;
```
ivarpruijn, line 40: Not the prettiest solution, but this fixes an error where this property is set in the DOM.

src/css/custom.css
```diff
@@ -115,12 +115,6 @@ body {
   .navbar .navSlack:hover {
     background: url('/img/icons/icon-nav-slack-hover.svg') no-repeat;
   }
-  .navbar .navDiscourse {
-    background: url('/img/icons/icon-nav-discourse.svg') no-repeat;
-  }
-  .navbar .navDiscourse:hover {
-    background: url('/img/icons/icon-nav-discourse-hover.svg') no-repeat;
-  }
```
ivarpruijn, line 123: Cleanup, no longer needed

docs/src/theme/NavbarItem/DropdownNavbarItem.tsx
```diff
@@ -0,0 +1,183 @@
...
+          if (e.key === 'Enter') {
+            e.preventDefault();
+            setShowDropdown(!showDropdown);
+          }
+        }}>
+        {props.children ?? props.label}
+      </NavbarNavLink>
+      {/* OBJECTIV: prevent collisions with active link on top of dropdown */}
+      <ul {...tagContent({ id: 'dropdown-menu' })} className="dropdown__menu">
+      {/* END OBJECTIV */}
```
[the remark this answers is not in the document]
sdirosa: Have you tried with tagChildren on the className?

docs/src/theme/NavbarItem/NavbarNavLink.tsx
ivarpruijn: Good question. Maybe because we use it in the site, which passes it on to the React tracker? But don't think we do that here in the browser tracker indeed.

sdirosa: There are two scenarios:
- If links in the configuration can be set as external with waitUntilTracked then this component is broken because it's not passing it correctly to tagLink. If that's the case we should also fix the Prop type definition, because as it is now it's incomplete.
- We actually don't have any external links in the docs and we should remove the waitUntilTracked options from the configuration.

docs/src/theme/NavbarItem/DropdownNavbarItem.tsx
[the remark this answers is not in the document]
ivarpruijn: just tried `tagChildren`, but whatever I do, collisions remain, even when they seemingly have complete different stacks like this one:
image

docs/src/theme/NavbarItem/NavbarNavLink.tsx
ivarpruijn: Just committed a fix, wdyt? https://github.com/objectiv/objectiv.io/pull/320/commits/9065db0a3dbd92cb63e1538e6a5a635a2aeadc60

docs/src/theme/NavbarItem/DropdownNavbarItem.tsx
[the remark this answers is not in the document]
sdirosa: If they really have completely different stacks then the error should not trigger and I've to investigate and fix this. It may be a bug in the Browser Tracker validation.

Could you please share your branch so I can investigate further?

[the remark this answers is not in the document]
ivarpruijn: Absolutely, made a branch with a (crude) test: `sidebar-dropdown-tagchildren`

docs/src/theme/NavbarItem/NavbarNavLink.tsx
sdirosa: Almost there! 

The `waitUntilTracked` option is still present in the docusaurus.config.json and gets passed onto the anchor tags as before. The error gets still thrown in console. 

I think this stuff is a leftover from a previous version of the components which could not perform isExternalLink  checks reliably because URLs where relative.

We should remove the waitUntilTracked options from docusaurus.config.json GitHub, Slack and Twitter items.

docs/src/theme/NavbarItem/NavbarNavLink.tsx
```diff
@@ -36,6 +36,8 @@ export default function NavbarNavLink({
...
   return (
     <Link
       {...tagLink({id: label as string, href: linkTo}) }
```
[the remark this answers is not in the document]
sdirosa: Let's say we add an icon to the text and this will break. I just wanted to point it out because it can be a pain to spot.

docs/src/theme/NavbarItem/NavbarNavLink.tsx
ivarpruijn: OK, that's my bad, I forgot to add the updated docusaurus.config.js that removes the `waitUntilTracked` parameters, sorry about that. So I did what you're saying, but didn't commit it 🤦 .

sdirosa: Ah! Then it's all good 👌🏼 
Let's just push that change and this works like a charm.

docs/src/theme/NavbarItem/DropdownNavbarItem.tsx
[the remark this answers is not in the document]
sdirosa: Thanks, will check this out asap! 🙏🏼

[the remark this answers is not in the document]
sdirosa: Found out why. This component clones the selected NavbarItem and renders it twice, once as the selected item and another time as a dropdown item. The validation only looks at the DOM, before Reacts fiddles with it, and sees two elements that are identical. Even though they are actually rendered in different locations eventually. 

Let's keep your hotfix for now, but I'll add this case in my TODO list to try and improve it. I think we can detect it correctly by refreshing the state of the validation after the DropdownItem has finished rendering.

It's kinda hard for BrowserTracker to detect these cases from the DOM alone, because this is React voodoo ;)

Of course an even better solution here would be to get rid of BrowserTracker and switch to ReactTracker for the docs, but I don't think we can until we have that SphinxPage stuff laying around.

Thanks for providing a reproducible example of the issue, really helped!

sdirosa:
@ivarpruijn No more comments from my side, I double checked everything. Only fix missing is that docusaurus.config change 👌🏼

ivarpruijn:
> @ivarpruijn No more comments from my side, I double checked everything. Only fix missing is that docusaurus.config change 👌🏼

Great! Just pushed that change, so will start testing this on staging today.

ivarpruijn:
Tested on staging, LGTM. Merging.

PR #321, opened 2022-03-04
'Join Slack' page

ivarpruijn:
We link in several places to join our Slack channel, via an invite link that is currently https://join.slack.com/t/objectiv-io/shared_invite/zt-u6xma89w-DLDvOB7pQer5QUs5B_~5pg.

Issues with this:
- Though we’ve set this link to never expire, there are [reports](https://www.reddit.com/r/Slack/comments/cmj56t/invite_link_set_to_never_expire_expired/) of that happening anyway.
- The link is used in at least 6 READMEs in our objectiv-analytics repo, which would all have to be updated if the invite expires.
- When it’s linked from anywhere else (e.g. our repo or on a forum) we cannot track that.
- The link doesn’t look very friendly.

This PR contains a PoC for a 'Join Slack' page (which will look like https://objectiv.io/join-slack), which redirects to the Slack invite link. If that link has to change, it only has to change in one place. Added benefits:

- The link looks a lot friendlier.
- We track this, so if it’s linked from somewhere other than our own site (e.g. a repo README or somewhere on a forum), we will also know.
- We can show a nice looking page while Slack is loading. In the future this could also be a different page where we tell more about our community first.

If we like this, we should make a bit of a design for the page (though it's only shown for a second before redirecting), which currently looks like this:
image

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -133,7 +133,7 @@ const config = {
         },
         {
           href: slackJoinLink,
-          label: ' ',
+          label: 'Slack',
           position: 'right',
           className: 'navItem navSlack',
           target: '_self',
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -7,7 +7,7 @@ const objectivEnvironment = process.env.OBJECTIV_ENVIRONMENT ?? 'development';
 const getEnvConfig = require('./env_config.js');
 const envConfig = getEnvConfig(objectivEnvironment);
 
-const slackJoinLink = 'https://join.slack.com/t/objectiv-io/shared_invite/zt-u6xma89w-DLDvOB7pQer5QUs5B_~5pg';
+const slackJoinLink = '/join-slack';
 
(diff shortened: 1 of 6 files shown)


docusaurus.config.js
```diff
@@ -133,7 +133,7 @@ const config = {
         },
         {
           href: slackJoinLink,
-          label: 'Slack',
+          label: ' ',
```
sdirosa, line 136: What's this change about?

sdirosa (approved):
Nice change, much easier to maintain 👌🏼

docusaurus.config.js
ivarpruijn: Sharp. Because this became an internal link, suddenly the text was showing, and this removes it. Though now I think of it another time, it would actually be a bit nicer to just hide it with CSS if we can, so I'll have a look at that.

sdirosa: Interesting, thanks for the explanation

PR #322, opened 2022-03-04
Move all CSS styling to their own pages

ivarpruijn:
We have CSS styling in two places:
- a `custom.css` file that is used on anything that Docusaurus outputs.
- one `styles.module.css` file that is currently used by all pages.

There are two issues with this setup:
1. Styling defined in `custom.css` can in rare case override the ones in `styles.module.css`, but only on a build. So you don't detect this when you run the site locally in dev mode. We've had this recently, with styling for `<h2>` defined in `styles.module.css` being overridden by `custom.css` in the build, which we of course only noticed when it already was on staging.
1. Styling changes in `styles.module.css` for one page can inadvertently change how another page looks.

This PR:
* Moves styling for each page to its own folder & `styles.module.css`, so they cannot interfere with each other.
* Moves overall styling as much as possible from `custom.css` to the separate `styles.module.css` files for each page.

To do:

- [ ] Test development version carefully, comparing to the production version step-by-step, for all breakpoints (desktop, tablet large, tablet medium, mobile). 
- [ ] Test staging version carefully, comparing to the production version step-by-step, for all breakpoints (desktop, tablet large, tablet medium, mobile).


jansenbob:
image

not sure if this is a result of the css rewrite, but there seems to be more margin between these two

ivarpruijn:
> image
> 
> not sure if this is a result of the css rewrite, but there seems to be more margin between these two

@jansenbob Not a result of this PR, but could use a tweak. I'll merge this PR as-is, and will do the tweak in another PR.

ivarpruijn:
Tested thoroughly on all breakpoints, desktop, and mobile - no differences found, so will merge.

PR #327, opened 2022-03-09
Update deploy commands in README

ivarpruijn:


README.md
```diff
@@ -63,7 +63,7 @@ docker run -e SFTP_URL \
     -e SFTP_USERNAME \
     -e SFTP_PASSWORD \
     -e SFTP_PUBKEY  \
-    objectiv/website-deploy deploy.sh
+    objectiv/website-deploy:$(date +%Y%m%d) deploy.sh
```
borft, line 66: so, if you don't set a tag (by setting TAG), when building the image, the default is the current date. Which is why this works. However, if you use something else, this will fail!

ivarpruijn: OK, so by default this works, right? I'll make a small note next to it with your comment.

borft: yes

borft (approved):

PR #330, opened 2022-03-11
Tweak styling for flex rows on mobile/tablet

ivarpruijn:
As found by @jansenbob in another PR. Still some pixels off due to how the flex box is rendered, but I think good enough.

Also added a small tweak to the README.


jansenbob (approved):
Good enough. ship it

PR #337, opened 2022-03-16
Diff sitemaps to detect changed URLs

ivarpruijn:
We have a growing number of references to site/docs URLs in various places. As an extra check to make sure we don't inadvertently break any URLs, this PR adds a step in 'Testing a branch on staging' to display the diff of URLs between what's on production and what's in the branch.


docker/build/xdiff.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/python3
+# xdiff.py -- compare XML files
+# Andreas Nolda 2022-02-07
+# https://hg.sr.ht/~nolda/xdiff
```
ivarpruijn, line 4: Found this tool online (GPL licensed), produces a nicely formatted XML diff. Not sure if it should be in this folder.

borft: not really, as it's not part of the docker image. But is there a particular reason we can't simply use `diff`? (which avoids using a custom script and python)

README.md
```diff
@@ -58,7 +58,20 @@ provision the appropriate environment variables for SFTP, and then run:
 # pull any changes, and build the deployable docker images for both website and docs for staging & production.
 git pull && make build-docker-build-image build-docker-deploy-image
 
-# then, upload to staging via FTP
+# second, compare the sitemap on production to the sitemap in the image, to double-check any URL changes; if 
+# you don't have `blessings` & `lxml` installed yet, do `sudo apt-get install python3-blessings python3-lxml`
+mkdir -p tmp
+curl https://objectiv.io/sitemap.xml > ./tmp/$(date +%Y%m%d)-sitemap-website.xml
```
borft, line 64: I'd put the date in a variable, and use that (which avoids the script breaking around midnight ;). Further more, put all the commands in a script, or simply in the `Makefile`

README.md
```diff
@@ -58,7 +58,20 @@ provision the appropriate environment variables for SFTP, and then run:
...
+docker run --rm --entrypoint cat objectiv/website-deploy:$(date +%Y%m%d) \
+  /services/build-production/sitemap.xml > ./tmp/$(date +%Y%m%d)-build-sitemap-website.xml
+curl https://objectiv.io/docs/sitemap.xml > ./tmp/$(date +%Y%m%d)-sitemap-docs.xml
+docker run --rm --entrypoint cat objectiv/website-deploy:$(date +%Y%m%d) \
+  /services/docs/build-production/sitemap.xml > ./tmp/$(date +%Y%m%d)-build-sitemap-docs.xml
+
+python3 xdiff.py ./tmp/$(date +%Y%m%d)-sitemap-docs.xml ./tmp/$(date +%Y%m%d)-build-sitemap-docs.xml
+python3 xdiff.py ./tmp/$(date +%Y%m%d)-sitemap-website.xml ./tmp/$(date +%Y%m%d)-build-sitemap-website.xml
+
```
borft, line 73: And maybe, when it is done, clean-up? btw, if you need temporary files, you could also use `mktemp`

README.md
```diff
@@ -58,7 +58,20 @@ provision the appropriate environment variables for SFTP, and then run:
...
 
-# then, upload to staging via FTP
+# second, compare the sitemap on production to the sitemap in the image, to double-check any URL changes; if 
+# you don't have `blessings` & `lxml` installed yet, do `sudo apt-get install python3-blessings python3-lxml`
+mkdir -p tmp
+curl https://objectiv.io/sitemap.xml > ./tmp/$(date +%Y%m%d)-sitemap-website.xml
+docker run --rm --entrypoint cat objectiv/website-deploy:$(date +%Y%m%d) \
+  /services/build-production/sitemap.xml > ./tmp/$(date +%Y%m%d)-build-sitemap-website.xml
+curl https://objectiv.io/docs/sitemap.xml > ./tmp/$(date +%Y%m%d)-sitemap-docs.xml
+docker run --rm --entrypoint cat objectiv/website-deploy:$(date +%Y%m%d) \
```
borft, line 68: I don't think you need to specify `--entrypoint`, you can simple use `cat <filename` as the command

README.md
```diff
@@ -58,7 +58,20 @@ provision the appropriate environment variables for SFTP, and then run:
...
+  /services/docs/build-production/sitemap.xml > ./tmp/$(date +%Y%m%d)-build-sitemap-docs.xml
+
+python3 xdiff.py ./tmp/$(date +%Y%m%d)-sitemap-docs.xml ./tmp/$(date +%Y%m%d)-build-sitemap-docs.xml
```
borft, line 71: btw, this relies on consistent ordering in the sitemaps, not sure if that is guaranteed.

docker/build/xdiff.py
ivarpruijn: Would be great if we could just use `diff`, but I tried it, and didn't get any useful (and nicely formatted) output out of that. I lack any skills here though, so if you have any tips LMK.

README.md
ivarpruijn: Thanks, will do.

README.md
ivarpruijn: Thanks, will do.

README.md
ivarpruijn: I think it's consistent from what I've seen so far, but will check Docusaurus' implementation to be sure.

README.md
ivarpruijn: Done

README.md
ivarpruijn: Added cleanup by adding `rm -rf /tmp` at the end. Tried `mktemp`, but didn't work for me, so left it for now.

README.md
ivarpruijn: Tried removing this, but didn't work

README.md
ivarpruijn: Checked the code, and pretty sure it's consistent, always sorted alphabetically. Also checked the resulting sitemap.xml files, and they're all sorted that way.

docker/build/xdiff.py
ivarpruijn: Tried that, but think it's a different package: https://pypi.org/project/xdiff/

ivarpruijn:
Doesn't affect any production code, so merging this now. No need for review @borft , we can discuss later.

PR #340, opened 2022-03-17
Track blog post header as ContentContext instead of NavigationContext

ivarpruijn:
Inline with the change we made on the homepage.

FYI @vincenthoogsteder

### DIFF
--- a/src/theme/BlogPostItem/index.tsx
+++ b/src/theme/BlogPostItem/index.tsx
@@ -115,7 +115,7 @@ function BlogPostArticle(props, blogPostId): JSX.Element {
 
   return (
     <>
-      <TrackedHeader>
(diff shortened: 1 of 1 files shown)


src/theme/BlogPostItem/index.tsx
```diff
@@ -115,7 +115,7 @@ function BlogPostArticle(props, blogPostId): JSX.Element {
 
   return (
     <>
-      <TrackedHeader>
+      <TrackedContentContext id='blog-post-header' Component='header'>
```
sdirosa, line 118: @ivarpruijn Let's have that chat about these TrackedComponents soon. 

I think that if we have the slightest doubt whether TrackedHeaders are navigation or content contexts, we should get rid of them. Same for TrackedFooters.

It's fairly easy to achieve the exact same as you did above here.

PR #343, opened 2022-03-19
Notification to 'star us' on scroll

ivarpruijn:
At a certain scroll position (currently at the 'Unlock a deeper level of insights without the typical gruntwork' section), this shows a small notification to 'star us', pointing to the top navigation. Only shows on desktop, not on tablet/mobile.


ivarpruijn:
@sdirosa Could you have a look when it's convenient if this is the right approach? Thanks!

PR #348, opened 2022-03-22
Docs: React Native small fixes

sdirosa:
This PR contains fixes around the examples in TrackedComponents for ReactNative.


sdirosa:
After a full review of the docs on staging, these are the only small fixes I found.

PR #351, opened 2022-03-23
Fix XSS issue sanitize-url

ivarpruijn:
See https://github.com/objectiv/objectiv.io/security/dependabot/20


sdirosa (approved):
Works locally for me

PR #354, opened 2022-03-23
Update dependency resolutions to fix Dependabot issues

ivarpruijn:
See https://github.com/objectiv/objectiv.io/security/dependabot

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -34,8 +34,9 @@
     "trim": "^0.0.3",
     "nth-check": "2.0.1",
     "node-fetch": "^3.2.0",
-    "node-forge": "^1.2.1",
+    "node-forge": "^1.3.0",
     "prismjs": "^1.27.0",
+    "minimist": "^1.2.6",
     "@braintree/sanitize-url": "^6.0.0"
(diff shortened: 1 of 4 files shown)


docs/package.json
```diff
@@ -34,8 +34,9 @@
     "trim": "^0.0.3",
     "nth-check": "2.0.1",
     "node-fetch": "^3.2.0",
-    "node-forge": "^1.2.1",
+    "node-forge": "^1.3.0",
     "prismjs": "^1.27.0",
+    "minimist": "^1.2.6",
     "@braintree/sanitize-url": "^6.0.0"
   },
```
sdirosa, line 41: 👌🏼

sdirosa (approved):
LGTM

docs/package.json
sdirosa: I guess at some point we should double check this resolution list, because probably some of them have been solved

ivarpruijn: Good point. I'll pick that up on the next Docusaurus upgrade that I'll do soon, will make a note.

PR #355, opened 2022-03-24
Improve search relevance

ivarpruijn:
- Makes certain words like 'React' optional for a search, so users get results without having to typo those.
- Provides more relevance to Tracker > Modeling > Taxonomy.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -165,7 +165,16 @@ const config = {
           // Public API key, safe to commit
         apiKey: '57e02ab0593f5338e36b7bff8235a505',
         indexName: 'objectiv',
-        container: 'main'
+        container: 'main',
+        contextualSearch: true,
+        searchParameters: {
+          optionalWords: ['react', 'react native', 'angular', 'browser'],
+          optionalFilters: [
+            'section:tracking<score=3>',
+            'section:modeling<score=2>',
+            'section:taxonomy<score=1>',
+          ],        
+        },
       },
     }),
 };


docs/docusaurus.config.js
```diff
@@ -165,7 +165,16 @@ const config = {
           // Public API key, safe to commit
         apiKey: '57e02ab0593f5338e36b7bff8235a505',
         indexName: 'objectiv',
-        container: 'main'
+        container: 'main',
+        contextualSearch: true,
```
ivarpruijn, line 169: Already enabled by default, just making it explicit here

docs/docusaurus.config.js
```diff
@@ -165,7 +165,16 @@ const config = {
...
+        searchParameters: {
+          optionalWords: ['react', 'react native', 'angular', 'browser'],
```
ivarpruijn, line 171: Makes certain keywords optional in a search, so users don't have to type them. We can add more.

docs/docusaurus.config.js
```diff
@@ -165,7 +165,16 @@ const config = {
...
+          optionalFilters: [
+            'section:tracking<score=3>',
+            'section:modeling<score=2>',
+            'section:taxonomy<score=1>',
```
ivarpruijn, line 175: Provides order of relevance to the various sections, so results in the Tracking section generally show up higher than Modeling, and then Taxonomy.

docs/docusaurus.config.js
sdirosa: Does this mean that if I search "Logging" or "Plugins" I would get results for any of the SDKs?

sdirosa (approved):
Looks good to me, left 1 question. 

Shall we test this on staging?

docs/docusaurus.config.js
ivarpruijn: Yes it would, but with the latest config changes I made they should now be grouped under their respective SDKs.

ivarpruijn:
> Looks good to me, left 1 question.
> 
> Shall we test this on staging?

Definitely, the Docusaurus beta 17 upgrade branch (which includes these changes) is on staging now. Already tested it yesterday, will do some more today.

sdirosa:
> > Looks good to me, left 1 question.
> > Shall we test this on staging?
> 
> Definitely, the Docusaurus beta 17 upgrade branch (which includes these changes) is on staging now. Already tested it yesterday, will do some more today.

Ah sorry, wasn't clear to me as the base branch of the upgrade branch wasn't this one. All good thanks!

ivarpruijn:
Merging, looks good on staging

docs/docusaurus.config.js
sdirosa: Very nice, I can see that on staging and I think it's a big improvement!

PR #357, opened 2022-03-24
Use built-in `useScrollPosition()` for the 'star us' notification

ivarpruijn:


src/components/star-us/index.tsx
```diff
@@ -22,19 +23,17 @@ function StarUsNotification(props, ref) {
...
-    };
-
-    document.addEventListener("scroll", onScroll);
-    return () => {
-      document.removeEventListener("scroll", onScroll);
-    };
-  }, [starUsAnchorRef, starUsNotificationClosed]);
+    },
+    [starUsNotificationClosed],
+  );
```
sdirosa, line 36: Very nice, can we get rid of "react-visibility-sensor" as well?
Doesn't seem to be used

sdirosa (approved):

src/components/star-us/index.tsx
ivarpruijn: Good one. I added that for the AnimatedGif component, but ended up taking a different approach. Will remove it in this PR.

PR #358, opened 2022-03-24
PoC: search metadata

ivarpruijn:
PoC to add metadata to certain docs to index with Algolia, and use in search results. Two parameters added:
- `platform`: for example 'React' or 'Angular'. Can be used to group search results, or in the future to filter on.
- `type`: for example 'Guide' or 'Core Concepts'. Can also be used to group search results, e.g. if not related to a specific platform.

Purely for discussion, let's not merge this yet.

### DIFF
--- a/docs/src/theme/SearchMetadata/index.tsx
+++ b/docs/src/theme/SearchMetadata/index.tsx
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+import Head from '@docusaurus/Head';
+import type {Props} from '@theme/SearchMetadata';
+// OBJECTIV
+interface PropsWithType extends Props {
+  type: string;
+  platform: string;
+}
+// END OBJECTIV
+
+// Note: we bias toward using Algolia metadata on purpose
+// Not doing so leads to confusion in the community,
+// as it requires to first crawl the site with the Algolia plugin enabled first
+// - https://github.com/facebook/docusaurus/issues/6693
+// - https://github.com/facebook/docusaurus/issues/4555
+export default function SearchMetadata({
+  locale,
+  version,
+  tag,
+  // OBJECTIV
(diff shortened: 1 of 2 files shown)


docs/src/theme/SearchMetadata/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      {/* Algolia DocSearch metadata */}
+      {language && <meta name="docsearch:language" content={language} />}
+      {version && <meta name="docsearch:version" content={version} />}
+      {tag && <meta name="docsearch:docusaurus_tag" content={tag} />}
+      {/* OBJECTIV */}
+      {type && <meta name="docsearch:type" content={type} />}
+      {platform && <meta name="docsearch:platform" content={platform} />}
+      {/* END OBJECTIV */}
+    </Head>
+  );
```
sdirosa, line 56: Why are we replacing the original implementation? Is it not allowed?
Can't we redefine just `type` and `platform` and leave the rest out?

ivarpruijn: I had to swizzle the component to be able to add these parameters. If we want to use this in production I agree that it's not necessary to swizzle this component; we can just add an own component (e.g. `CustomSearchMetaData`) to add `type` and `platform` search metadata.

sdirosa: Why did you need to swizzle it?

Can't we just create something like this?

```tsx
import Head from '@docusaurus/Head';

type CustomSearchMetadataProps = {
  type: string, // can be used for grouping search results in the future
  platform: string // can be used for filtering on the platform type (e.g. 'React') in the future
}

export const CustomSearchMetadata = ({ type, platform }: CustomSearchMetadataProps) => (
  <Head>
    {type && <meta name="docsearch:type" content={type} />}
    {platform && <meta name="docsearch:platform" content={platform} />}
  </Head>
);
```

ivarpruijn: Yes, that's exactly what we need if we want to use this in production, instead of swizzling the component. I just swizzled it to test if this could work, this is not to be merged.

ivarpruijn:
Closing this PR, will work on a new one to add optional custom search metadata

PR #360, opened 2022-03-25
Scan repos for removed URLs from docs (if any)

ivarpruijn:
Adds a script to check the objectiv-analytics and objectiv.io repos for any URLs that have been removed from the docs. It specifically scans only `.{md,mdx,rst,html,ipynb}`  files. Script is called directly from diff-sitemaps.sh.


docker/build/scan-urls-repo.sh
```diff
@@ -0,0 +1,40 @@
...
+      tput setaf 2
+      echo "No removed URLs found in $DIRECTORY"
+    else
+      tput setaf 1
+      echo $rec1_no_cr "is used in:"
+      echo "$OUTPUT"
+    fi
+  done < <(cut -d "," -f1,3 $PWD/$URLSFILE | tail -n +2)
+else
+  echo "No file '$URLSFILE' to scan, aborting"
```
borft, line 39: don't think `$URLSFILE` will be printed here because of the single quotes

docker/build/diff-sitemaps.sh
```diff
@@ -1,13 +1,45 @@
...
+      ;;
+  esac
+
+done
+
 mkdir -p ./tmp
 curl https://objectiv.io/sitemap.xml > ./tmp/$CURRENT_DATE-sitemap-website.xml
 docker run --rm --entrypoint cat objectiv/website-deploy:$CURRENT_DATE \
-  /services/build-production/sitemap.xml > ./tmp/$CURRENT_DATE-build-sitemap-website.xml
+  /services/build-$ENVIRONMENT/sitemap.xml > ./tmp/$CURRENT_DATE-build-sitemap-website.xml
```
ivarpruijn, line 32: Can now specify the environment, e.g. `testing`. By default it's `production`.

docker/build/diff-sitemaps.sh
```diff
@@ -1,13 +1,45 @@
...
 
 python3 ./docker/build/xdiff.py ./tmp/$CURRENT_DATE-sitemap-docs.xml ./tmp/$CURRENT_DATE-build-sitemap-docs.xml
 python3 ./docker/build/xdiff.py ./tmp/$CURRENT_DATE-sitemap-website.xml ./tmp/$CURRENT_DATE-build-sitemap-website.xml
 
-rm -rf ./tmp/
\ No newline at end of file
+# check if any URLs that are removed appear in the objectiv-analytics repo
+./docker/build/scan-urls-repo.sh -d=../objectiv-analytics -u=./tmp/removed_urls.csv
+# check if any URLs that are removed appear in the objectiv.io repo
+./docker/build/scan-urls-repo.sh -d=./ -u=./tmp/removed_urls.csv
```
ivarpruijn, line 43: Scan both objectiv-analytics and objectiv.io repos

docker/build/xdiff.py
```diff
@@ -120,43 +122,120 @@ def diff_strings(string1, string2, file1, file2):
...
+    filedata = filedata.replace(old_text, new_text)
+
+    # Write the file out again
+    with open(file_to_replace, 'w') as file:
+        file.write(filedata)
+
+def main():
+    # replace certain environment lines in the file and write it back
+    replace_text_in_file(args.file1, "testing.objectiv.io", "objectiv.io")
+    replace_text_in_file(args.file2, "testing.objectiv.io", "objectiv.io")
```
ivarpruijn, line 216: Make it work with the testing environment

README.md
```diff
@@ -75,7 +77,7 @@ git pull && make build-docker-build-image build-docker-deploy-image
 
 # second, compare the sitemap on production to the sitemap in the image, to double-check any URL changes; if 
 # you don't have `blessings` & `lxml` installed yet, do `sudo apt-get install python3-blessings python3-lxml`
-./docker/build/diff-sitemaps.sh
+sh ./docker/build/diff-sitemaps.sh
```
borft, line 80: you call it using `sh`, but the script actually has a `#!/bin/bash`. Can't we just make the file executable (chmod +x), so it gets run in the correct shell?

docker/build/diff-sitemaps.sh
```diff
@@ -1,13 +1,45 @@
+#!/bin/bash
+
+
 CURRENT_DATE=$(date +%Y%m%d)
+ENVIRONMENT=production
```
borft, line 5:
```suggestion
if [ -z "$ENVIRONMENT"];
then
  ENVIRONMENT=production
fi 
```

how about only setting a default if it's not set. That way it behaves the same as the `Makefile`.

borft: actually, maybe we could rename it to `OBJECTIV_ENVIRONMENT`

docker/build/diff-sitemaps.sh
```diff
@@ -1,13 +1,45 @@
...
+
 mkdir -p ./tmp
 curl https://objectiv.io/sitemap.xml > ./tmp/$CURRENT_DATE-sitemap-website.xml
 docker run --rm --entrypoint cat objectiv/website-deploy:$CURRENT_DATE \
-  /services/build-production/sitemap.xml > ./tmp/$CURRENT_DATE-build-sitemap-website.xml
+  /services/build-$ENVIRONMENT/sitemap.xml > ./tmp/$CURRENT_DATE-build-sitemap-website.xml
 curl https://objectiv.io/docs/sitemap.xml > ./tmp/$CURRENT_DATE-sitemap-docs.xml
 docker run --rm --entrypoint cat objectiv/website-deploy:$CURRENT_DATE \
-  /services/docs/build-production/sitemap.xml > ./tmp/$CURRENT_DATE-build-sitemap-docs.xml
+  /services/docs/build-$ENVIRONMENT/sitemap.xml > ./tmp/$CURRENT_DATE-build-sitemap-docs.xml
```
borft, line 35: maybe we can use actual temp files using `mktemp` here. So it doesn't break if we run the script in parallel

docker/build/scan-urls-repo.sh
```diff
@@ -0,0 +1,44 @@
...
+    -*|--*)
+      echo "Unknown option $i"
+      exit 1
+      ;;
+    *)
+      ;;
+  esac
+done
+
+if test -f "$URLSFILE"; then
```
borft, line 22:
```suggestion
if [ -f "$URLSFILE" ]; then
```
consistency

docker/build/scan-urls-repo.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "Scanning directory '$DIRECTORY' from URLs file '$URLSFILE'...";
+  FOUND_URLS=0
```
borft, line 24:
```suggestion
declare -i  FOUND_URLS=0
```

make it a proper integer

docker/build/scan-urls-repo.sh
```diff
@@ -0,0 +1,44 @@
...
+    OUTPUT=$(grep --include=\*.{md,mdx,rst,html,ipynb} --exclude-dir={node_modules,tests} -rnl $DIRECTORY -e $rec1_no_cr)
+    if ! [ -z "$OUTPUT" ]
+    then
+      let FOUND_URLS++
+      tput setaf 1
+      echo $rec1_no_cr "is used in:"
+      echo "$OUTPUT"
+    fi
+  done < <(cut -d "," -f1,3 $PWD/$URLSFILE | tail -n +2)
+  if [ $FOUND_URLS == 0 ];
```
borft, line 37:
```suggestion
  if [ $FOUND_URLS -eq 0 ];
```
numerical comparison

docker/build/xdiff.py
borft: oh dear, so we actually change the files we're comparing during the comparison?

docker/build/xdiff.py
```diff
@@ -9,6 +9,8 @@
 from blessings import Terminal
```
borft, line 9: do we really need a custom script to do a simple diff?

docker/build/xdiff.py
```diff
@@ -120,43 +122,120 @@ def diff_strings(string1, string2, file1, file2):
...
+def replace_text_in_file(file_to_replace, old_text, new_text):
+    # Read in the file
+    with open(file_to_replace, 'r') as file:
+        filedata = file.read()
+
+    # Replace the target string
+    filedata = filedata.replace(old_text, new_text)
+
+    # Write the file out again
+    with open(file_to_replace, 'w') as file:
```
borft, line 210: why do we open the same file twice?

README.md
```diff
@@ -75,7 +77,7 @@ git pull && make build-docker-build-image build-docker-deploy-image
 
 # second, compare the sitemap on production to the sitemap in the image, to double-check any URL changes; if 
 # you don't have `blessings` & `lxml` installed yet, do `sudo apt-get install python3-blessings python3-lxml`
```
borft, line 79: can we use a virtualenv for this please? also `lxml` is a different package then `python3-lxml`

docker/build/xdiff.py
```diff
@@ -9,6 +9,8 @@
 from blessings import Terminal
 from lxml import etree
```
borft, line 10: why do we need lxml?

docker/build/scan-urls-repo.sh
```diff
@@ -0,0 +1,44 @@
...
+done
+
+if test -f "$URLSFILE"; then
+  echo "Scanning directory '$DIRECTORY' from URLs file '$URLSFILE'...";
+  FOUND_URLS=0
+  while IFS="," read -r rec1
+  do
+    rec1_no_cr="${rec1/$'\r'/}"
+    OUTPUT=$(grep --include=\*.{md,mdx,rst,html,ipynb} --exclude-dir={node_modules,tests} -rnl $DIRECTORY -e $rec1_no_cr)
+    if ! [ -z "$OUTPUT" ]
```
borft, line 29:
```suggestion
    if [ -n "$OUTPUT" ]
```

docker/build/xdiff.py
ivarpruijn: Once to read, then to write. But indeed inefficient, will make it one statement.

docker/build/xdiff.py
ivarpruijn: Couldn't find a way to do it in memory (though exactly the same effect), so did it like this. Will fix when we're making this a simpler script.

docker/build/scan-urls-repo.sh
ivarpruijn: Works on my machine actually

borft: yeah, you're right, it's because it's enclosed in the double quotes.
```
mike@mikepad ~ → URLSFILE=test
mike@mikepad ~ → echo '$URLSFILE'
$URLSFILE
mike@mikepad ~ → echo "'$URLSFILE'"
'test'
```

docker/build/diff-sitemaps.sh
borft: something along the lines of:
```
build_sitemap_docs=$(mktemp)
docker run ..... > $build_sitemap_docs
....
em $build_sitemap_docs
```

docker/build/xdiff.py
```diff
@@ -120,43 +122,122 @@ def diff_strings(string1, string2, file1, file2):
...
+    # print and write removed URLs
+    if(len(parsed_diff[1]) > 0):
+        removed_urls = extract_urls_from_diff(parsed_diff[1])
+        print(term.red("Removed URLs:"), end="\n")
+        for url in removed_urls:
+            for u in url:
+                print(term.red("- "+u), end="\n")
+        write_urls_to_csv(urls=removed_urls, file='./tmp/removed_urls.csv')
+    
+    return 1
```
borft, line 240: do we always return 1? This implies we always fail

docker/build/xdiff.py
```diff
@@ -120,43 +122,122 @@ def diff_strings(string1, string2, file1, file2):
...
+    with open(file_to_replace, 'r+') as file:
+        filedata = file.read()
+        # Replace the target string
+        filedata = filedata.replace(old_text, new_text)
+        # Setting the position to the top
+        # of the page to insert data
+        file.seek(0)
+        # Write the file out again
+        file.write(filedata)
+        # Truncating the file size
```
borft, line 212: shouldn't we truncate first, and then write the data?

docker/build/xdiff.py
```diff
@@ -120,43 +122,122 @@ def diff_strings(string1, string2, file1, file2):
...
+          search = re.search("<loc>(.+?)</loc>", str)
+          if (search):
+              url = search.group(1)
+              parsed_url = url.replace('testing.objectiv.io', 'objectiv.io')
+              parsed_url = parsed_url.replace('staging.objectiv.io', 'objectiv.io')
+              extracted_urls.append([parsed_url])
+      
+  return extracted_urls
+
+def write_urls_to_csv(urls, file='./tmp/urls.csv'):
```
borft, line 187: what if there's no `./tmp` in the current directory? wouldn't it be better to use `/tmp`?

docker/build/xdiff.py
```diff
@@ -120,43 +122,122 @@ def diff_strings(string1, string2, file1, file2):
...
+
+def main():
+    # replace certain environment lines in the file and write it back
+    replace_text_in_file(args.file1, "testing.objectiv.io", "objectiv.io")
+    replace_text_in_file(args.file2, "testing.objectiv.io", "objectiv.io")
+
+    parsed_diff = parse_diff(args)
+
+    # print and write added URLs
+    if(len(parsed_diff[0]) > 0):
```
borft, line 223: empty List evaluates to `False`
```suggestion
    if parsed_diff[0]:
```

docker/build/xdiff.py
borft: hmm, can we make the match stricter? maybe replace https://testing.objectiv.io/ with https://objectiv.io/ or something like that, to lower the chance of accidentally replacing something we shouldn't?

ivarpruijn:
Closing, replaced by https://github.com/objectiv/objectiv.io/pull/376

PR #361, opened 2022-03-26
Testing environment build & deploy

ivarpruijn:
Adds the ability to build & deploy to a test environment (testing.objectiv.io).

Todo:
- [x] Make deploy script work  (currently errors on `/usr/local/bin/docker-entrypoint.sh: 11: exec: deploytesting.sh: not found`)


Makefile
```diff
@@ -27,6 +27,10 @@ build-docker-build-image:
 	$(info building for ${OBJECTIV_ENVIRONMENT})
 	docker build --build-arg OBJECTIV_ENVIRONMENT=$(OBJECTIV_ENVIRONMENT) --no-cache -t objectiv/website-build:${TAG} -f docker/build/Dockerfile .
 
+build-docker-test-image: OBJECTIV_ENVIRONMENT="testing"
```
borft, line 30: can't we simply do:
```bash
make OBJECTIV_ENVIRONMENT=testing build-docker-build-image
```?

File: .htaccess-testing [hunk 2]
```diff
@@ -0,0 +1,15 @@
+# redirect to https
+RewriteEngine On
+RewriteCond %{HTTP_HOST} ^objectiv.io [NC]
+RewriteCond %{HTTP:X-Forwarded-Proto} !https
+RewriteCond %{HTTPS} off
+RewriteRule ^(.*)$ https://objectiv.io/$1 [R,L]
```
borft, line 6: don't think this will do a lot

ivarpruijn: As discussed, will leave it here for now, we can update this in the future if needed for e.g. adding a .htpasswd

Makefile
ivarpruijn: That should work, will update

PR #367, opened 2022-03-31
Use categories in model hub restructure

ivarpruijn:
This PR uses categories for the model hub restructure where applicable.

Noticeable other changes:
- No longer opens the first page under a category immediately; resulted in some strange behavior here and there, and is no longer really needed with the new category pages.
- Makes headings for the API references only a bit smaller.

### DIFF
--- a/docs/docs/modeling/DataFrame.mdx
+++ b/docs/docs/modeling/DataFrame.mdx
@@ -7,8 +7,8 @@ slug: /modeling/DataFrame
 
 ---
 <head>
-    <meta name="description" content="A mutable DataFrame representing tabular data in a database and enabling operations on that data." />
-    <meta property="og:description" content="A mutable DataFrame representing tabular data in a database and enabling operations on that data." />
+    <meta name="description" content="DataFrame — Bach  documentation" />
+    <meta property="og:description" content="DataFrame — Bach  documentation" />
 </head>
 
 export const toc = [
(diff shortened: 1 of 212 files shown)


docs/src/css/custom.css
```diff
@@ -301,12 +301,19 @@ div.docs-wrapper footer div.col:nth-child(2) {
 /** DOCS - Modeling **/
 
 /* Make API reference titles a bit smaller */
-h1[id^=bach], h1[id^=modelhub] {
+h1[id^=bach-dataframe], h1[id^=bach-series], h1[id^=modelhub] {
```
ivarpruijn, line 304: Make API reference titles (only) a bit smaller

docs/src/theme/DocSidebarItem/index.tsx
```diff
@@ -195,9 +185,6 @@ function DocSidebarItemCategory({
                   if (href) {
                     updateCollapsed(false);
                   } else {
-                    // OBJECTIV: auto-select the first item in the category if it's a link
-                    firstItemInCategoryLink != "" ? history.push(firstItemInCategoryLink) : null;
-                    // OBJECTIV END
```
ivarpruijn, line 200: No longer auto-select the first item in a category, to avoid strange UX esp. with the 'Reference by function' categories.

docs/docs/modeling/DataFrame.mdx
```diff
@@ -7,8 +7,8 @@ slug: /modeling/DataFrame
 
 ---
 <head>
-    <meta name="description" content="A mutable DataFrame representing tabular data in a database and enabling operations on that data." />
-    <meta property="og:description" content="A mutable DataFrame representing tabular data in a database and enabling operations on that data." />
+    <meta name="description" content="DataFrame — Bach  documentation" />
+    <meta property="og:description" content="DataFrame — Bach  documentation" />
```
ivarpruijn, line 11: Would like to fix this, but doesn't seem like a blocker. Shall we fix this for the next release @jansentom ?

ivarpruijn (commented):
@jansentom A few comments to look at, but overall just ignore the rest, esp. all the auto-generated .mdx and .html files.

docs/docs/modeling/DataFrame.mdx
ivarpruijn: Fix committed

PR #377, opened 2022-04-05
Detect non-absolute URLs

ivarpruijn:
Our broken-link-checker cannot follow non-absolute URLs. This adds a check to the check_urls script for usage of non-absolute URLs in the website & blog.

This PR also updates:
- The README file to create a virtualenv for running the script. 
- A small error in a blog post, found accidentally by the script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,13 @@ provision the appropriate environment variables for SFTP, and then run:
 git pull && make build-docker-test-image build-docker-deploy-image
 
 # check image for any removed/changed URLs that may cause issues
+cd scripts
+virtualenv check_sitemaps
+source check_sitemaps/bin/activate
+pip install -r requirements.txt
+cd ../
 python3 scripts/check_urls.py --environment=testing
+deactivate
 
 # then, upload to testing via FTP
 docker run -e SFTP_URL \
@@ -77,7 +83,13 @@ provision the appropriate environment variables for SFTP, and then run:
 git pull && make build-docker-build-image build-docker-deploy-image
 
 # check image for any removed/changed URLs that may cause issues
+cd scripts
+virtualenv check_sitemaps
+source check_sitemaps/bin/activate
+pip install -r requirements.txt
+cd ../
 python3 scripts/check_urls.py --environment=staging
+deactivate
 
 # if URL checks are okay, upload to staging via FTP
 docker run -e SFTP_URL \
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -58,7 +58,13 @@ provision the appropriate environment variables for SFTP, and then run:
 git pull && make build-docker-test-image build-docker-deploy-image
 
 # check image for any removed/changed URLs that may cause issues
+cd scripts
+virtualenv check_sitemaps
+source check_sitemaps/bin/activate
+pip install -r requirements.txt
+cd ../
 python3 scripts/check_urls.py --environment=testing
+deactivate
```
ivarpruijn, line 67: Create a virtualenv first

blog/2022-02-16 Location Stack.md
```diff
@@ -67,7 +67,7 @@ Here is an example of the same event, but this time collected by Objectiv’s tr
    {
      "_type": "LinkContext",
      "id": "overview",
-     "href": "/docs/modeling/DataFrame""
+     "href": "https://objectiv.io/docs/modeling/DataFrame"
```
ivarpruijn, line 70: Double quotation mark, but should also be a full URL

scripts/check_urls.py
```diff
@@ -17,6 +18,36 @@
...
+    """
+
+    non_abs_urls_found_in_files = []
+    for filename in glob.glob(path, recursive=True):
+        if has_extension(filename, extensions):
+            with open(filename) as f:
+                contents = f.read()
+                for url in urls:
+                    # find all occurences of the URL and determine if it's non-absolute
+                    matches = re.findall("(.*?)\/" + url, contents)
```
borft, line 41:
```suggestion
                    matches = re.findall(f'(.*?)\/{url}', contents)

```

use f-string and single quotes instead of concatenation

README.md
ivarpruijn: Ofc, will update that

scripts/check_urls.py
```diff
@@ -17,6 +18,36 @@
...
+                    for match in matches:
+                        before = match.strip()
+                        # non-absolute if: doesn't use `useBaseUrl()` or contains "https://" or Tracking code
+                        # could do more specific URL matching here, but sufficient for this purpose
+                        if (before.find("useBaseUrl") == -1 and before.find("https://") == -1
```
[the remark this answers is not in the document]
ivarpruijn: Good one, will also check for `http://`.

We could check the build artefacts, but then I'd have to extract those from the build first, and figured that would a bit messy working with temp files etc. So instead I went for scanning what's currently on disk.

PR #386, opened 2022-04-15
Fix build warning by regenerating yarn.lock

ivarpruijn:
There was a build warning from webpack that regularly popped up. Looked around for it, but didn't find anyone with the same issue. Fully deleting and regenerating yarn.lock fixed it, so I think it was just a conflict in the lockfile.

This PR replaces the old yarn.lock files with the regenerated yarn.lock files. Tested and all pages still work.


sdirosa (approved):
Very nice, I've been having this issue for weeks.

PR #394, opened 2022-04-22
Fix VisibleEvent having no ContentContext on blog post load

ivarpruijn:
This PR fixes the issue with the ContentContext no longer being present for a blog post load.

It also include a small tweak to the looks of the header with the avatar.


vincenthoogsteder (approved):
LGTM

PR #404, opened 2022-05-06
Add stars count to homepage

ivarpruijn:
Shows the repo's stars count on the homepage, using GitHub's REST API. Using unauthenticated requests, which should be fine, as the rate limit is 60 per hour per IP address (based on the user's IP). If rate limit is exceeded or any other error occurs, the count is not shown at all.

Also fixes an unrelated warning about `<h3>` tags used within `<p>` tags.


vincenthoogsteder (approved):
LGTM

PR #412, opened 2022-05-13
Move stars counters into button

ivarpruijn:
Moves the stars counter into the button, instead of next to it. Now looks like this:

image

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -199,7 +199,7 @@ export default function Home() {
                   <p>
                     Objectiv&nbsp;
                     <TrackedLink
-                      to={useBaseUrl("/docs/modeling/example-notebooks/", {absolute: true})}
+                      to={useBaseUrl("/docs/modeling/open-model-hub/", {absolute: true})}
                       waitUntilTracked={true}
                       target="_self">
                       includes pre-built models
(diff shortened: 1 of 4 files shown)


jansenbob:
All LGTM, but i'm not a big fan of the blue hover color:

image

Maybe we can just leave the button background color as is, as the drop shadow already indicates that you're hovering the button.

jansenbob (changes requested):
change blue background on hover to FAFAFA

src/pages/index.tsx
```diff
@@ -199,7 +199,7 @@ export default function Home() {
                   <p>
                     Objectiv&nbsp;
                     <TrackedLink
-                      to={useBaseUrl("/docs/modeling/example-notebooks/", {absolute: true})}
+                      to={useBaseUrl("/docs/modeling/open-model-hub/", {absolute: true})}
```
ivarpruijn, line 202: @jansenbob @vincenthoogsteder Updated the link on the homepage to take you directly to the open model hub docs. Thought that made more sense than the example notebooks or the introduction page of the modeling section. WDYT?

vincenthoogsteder: agreed! 

shall we also do that with the GH readme then? That now points to modeling intro, but agree this is better

jansenbob (approved):
LGTM!

src/pages/index.tsx
ivarpruijn: let's!

PR #416, opened 2022-05-16
Add GZIP compression

ivarpruijn:
Enables GZIP compression for the site + docs. 

Tested on testing environment and is returning gzip encoding:
gzip
t


.htaccess-production
```diff
@@ -1,3 +1,47 @@
+<IfModule mod_deflate.c>
```
borft, line 1: maybe include the link to transip's docs here for reference

borft (approved):
lgtm, one small suggestion

.htaccess-production
ivarpruijn: good one, done

PR #419, opened 2022-05-16
Use latest build of the tracker

borft:


docs/package.json
```diff
@@ -17,7 +17,7 @@
     "@docusaurus/preset-classic": "2.0.0-beta.18",
```
borft, line 17: Nice catch @ivarpruijn

ivarpruijn (approved):
Tested locally and all events on site & docs are still working. Let's merge!

PR #420, opened 2022-05-16
Fix tracking warning on empty `PaginatorNavLink` title

ivarpruijn:
If the title for the link in the `PaginatorNavLink` was empty, e.g. if the previous page has an empty title, the Tracker is not able to tag the link and throws a warning. This happens for instance on the first two pages of docs home.

### DIFF
--- a/docs/src/theme/PaginatorNavLink/index.tsx
+++ b/docs/src/theme/PaginatorNavLink/index.tsx
@@ -20,7 +20,7 @@ export default function PaginatorNavLink(props: Props): JSX.Element {
       className="pagination-nav__link" 
       to={permalink}
       // OBJECTIV
-      {...tagLink({ id: title as string, href: permalink })}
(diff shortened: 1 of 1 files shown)


docs/src/theme/PaginatorNavLink/index.tsx
```diff
@@ -20,7 +20,7 @@ export default function PaginatorNavLink(props: Props): JSX.Element {
       className="pagination-nav__link" 
       to={permalink}
       // OBJECTIV
-      {...tagLink({ id: title as string, href: permalink })}
+      {...tagLink({ id: (title as string != "" ? title as string : "[no title]"), href: permalink })}
```
borft, line 23: could we do something like:
id: (title ?? '[no title]')

ivarpruijn: Unfortunately no, the title is always defined/not null, it's just sometimes an empty string.

Issue #426, opened 2022-05-19
Browser tracker >= 0.0.18 crash on /docs


ivarpruijn:
**Describe the bug**
When accepting cookies on /docs, the browser tracker crashes with an error `Tracker 'objectiv-docs-dev' exists but its configuration doesn't match the given one.`. After refresh the docs and the tracker function normally. 

**To Reproduce**
1. Go to /docs in an incognito window
2. Accept the cookies
3. See the error

**Expected behavior**
The tracker should initialize correctly without any errors.

**Additional information**
Issue occurs only on version >=0.0.18.

PR #427, opened 2022-05-19
Avoid re-creating the Tracker instance on re-render in Root.tsx

sdirosa:
@ivarpruijn This is the fix. Basically we should never re-create the Tracker. 
In this particular case the issue has been introduced because the configuration changes (the active bit is false on creation and true on re-render). I think we can improve this by ignoring the active bit, since that's something that can change at runtime.

Another issue is the custom plugin being an anonymous function. By definition that will change every time and cannot be compared.

I'll see if I can improve on both of these in the next version of the Browser Tracker. I've a couple of ideas.

Fixes https://github.com/objectiv/objectiv.io/issues/426


ivarpruijn:
Tested on the testing environment and LGTM

PR #442, opened 2022-06-02
Announcement bar for the Launchpad

ivarpruijn:
Adds a bar on top to announce the Launchpad.
- Links to the Launchpad docs.
- Currently only on the homepage. @jansenbob @vincenthoogsteder Shall we add it to the other pages as well?


vincenthoogsteder:
@ivarpruijn think only on home is fine

jansenbob (approved):
LGTM

PR #456, opened 2022-06-10
Trackers v0.0.20

sdirosa:
Changes:

- Added .yarnrc to .gitignore of both website and docs
- Upgraded React Tracker in both website and docs to v0.0.20
- Website
  - Modified Root.tsx to load the developer tools
  - Fixed Success and Failure events to use the new APIs
- Docs
  - Modified Root.tsx to load the developer tools 
  - Fixed `PaginatorNavLinkWrapper` component: missing href obligatory attribute
 
**NOTE: There are MANY other TypeScript violations in both projects, but I don't think fixing those is in the scope of this PR.**

### Publishing 

Until v0.0.20 is published to NPM, these changes can be only built with Verdaccio:

- Start up Verdaccio if not already running
- Checkout this branch https://github.com/objectiv/objectiv-analytics/pull/873
- In the tracker folder run `yarn install && yarn build`
- If all went ok, test publishing to verdaccio with `TAG=next yarn publish:verdaccio`
- Check if publishing worked with http://localhost:4873/
- If all ok, reset verdaccio data store and publish again with `TAG=latest yarn publish:verdaccio`
- Add .yarnrc containing `"@objectiv:registry" "http://127.0.0.1:4873"` to both website and docs
- Do a yarn install in both website and docs, it should pull v0.0.20 from Verdaccio
- Build website and docs with v0.0.20


.gitignore
```diff
@@ -4,6 +4,9 @@
 # Production
 /build
 
+# Location Yarn configuration
+.yarnrc
+
```
sdirosa, line 9: This is used for Verdaccio only but should never be commited.

docs/.gitignore
```diff
@@ -7,6 +7,9 @@
 # Yarn Integrity file
 .yarn-integrity
 
+# Location Yarn configuration
+.yarnrc
+
```
sdirosa, line 12: This is used for Verdaccio only but should never be commited.

docs/src/theme/PaginatorNavLink/index.tsx
```diff
@@ -11,6 +11,7 @@ export default function PaginatorNavLinkWrapper(props: Props): JSX.Element {
   return (
     <TrackedLinkContext 
       Component={'div'}
+      href={props.permalink}
```
sdirosa, line 14: Bug: Both TS and Collector rejected this event, because LinkContexts must provide href.

docs/src/theme/Root.tsx
```diff
@@ -36,6 +35,7 @@ const cookiebotConsentStatistics = (): boolean => {
 }
 
 function Root({children}) {
+  const [tracker, setTracker] = useState<ReactTracker>();
```
sdirosa, line 38: Latest Docusaurus seems to be re-rendering this Root component on every routing operation. This caused the Tracker to be re-created on every link click, like the header menu.

I solved the issue by creating a state variable to hold the Tracker instance and creating it only once, if not already set. Below at line 50.

src/components/github-stargazers/index.tsx
```diff
@@ -58,7 +56,9 @@ function Stargazers(cta) {
         console.log(err.response?.data || err);
         setStargazersCount(null);
         // track the error
-        trackFailureEvent()
+        trackFailureEvent({
+          message: "Could not retrieve stargazers count"
+        })
       },
```
sdirosa, line 62: Changed this event tracking call to latest API. The message parameter moved from the hook factory to the callback.

src/pages/join-slack/index.tsx
```diff
@@ -9,10 +9,10 @@ export default function JoinSlack() {
...
     'https://join.slack.com/t/objectiv-io/shared_invite/zt-19agzjvbn-jfIPqrYUOlrqeun2fZakCw'
     // 'https://join.slack.com/t/objectiv-io/shared_invite/zt-u6xma89w-DLDvOB7pQer5QUs5B_~5pg';
-  const trackSuccessEvent = useSuccessEventTracker({message: 'Redirect to Slack'});
+  const trackSuccessEvent = useSuccessEventTracker();
 
   useEffect(() => {
-    trackSuccessEvent();
+    trackSuccessEvent({message: 'Redirect to Slack'});
     window.location.href = slackJoinLink;
   }, []);
```
sdirosa, line 17: Changed this event tracking call to latest API. The message parameter moved from the hook factory to the callback.

src/components/github-stargazers/index.tsx
ivarpruijn: 👍

docs/src/theme/PaginatorNavLink/index.tsx
ivarpruijn: Nice catch, thanks

docs/src/theme/Root.tsx
ivarpruijn: looks good

ivarpruijn (approved):
Tested & LGTM. Found one unrelated collision in the blog, which I'll fix separately

PR #458, opened 2022-06-10
Fix collision in blog images

ivarpruijn:
If a blog image has no caption, and there are more than one of them, a collision occurs, such as is the case for the latest blog post. This PR fixes that.

### DIFF
--- a/src/components/blog-image/index.js
+++ b/src/components/blog-image/index.js
@@ -38,6 +38,9 @@ function BlogImage({
     classNames = clsx(styles.blogImage, styles.blogImageLarge)
   }
 
+  // ensure the tracked div has an ID even if there's no imageAlt (with links)
+  imageAlt = (imageAlt == '') ? url : imageAlt; 
+
   return (
     <TrackedDiv id={'image: ' + imageAlt}>
       <img src={useBaseUrl(url)} alt={imageAlt} className={classNames} />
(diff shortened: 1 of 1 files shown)


src/components/blog-image/index.js
```diff
@@ -38,6 +38,9 @@ function BlogImage({
     classNames = clsx(styles.blogImage, styles.blogImageLarge)
   }
 
+  // ensure the tracked div has an ID even if there's no imageAlt (with links)
+  imageAlt = (imageAlt == '') ? url : imageAlt;
```
sdirosa, line 42: You can simplify as 
```
imageAlt = imageAlt ?? url; 
```

sdirosa (approved):
Looks good, left a small simplification suggestion

src/components/blog-image/index.js
ivarpruijn: `imageAlt` is an empty string by default, so don't think this works

sdirosa: If it's an empty string by default, then the url is going to be used. Isn't that the same of what the code is checking now?

The only difference in my approach is that now it works also for null and undefined.

sdirosa: Oh wait I get it now, so that variable is actually equal to '' by default?

Then I'd add the `===` instead of `==`.

PR #461, opened 2022-06-15
Website: e2e testing

sdirosa:
In this PR we setup cypress to test the objectiv instrumentation of the website.

Here is a recap of what's been done.

## Upgraded to a version of React Tracker that has EventRecorder
```
yarn remove @objectiv/tracker-react
yarn add @objectiv/tracker-react@next
```

## Installed Cypress with the snapshot plugin
```
yarn add cypress @cypress/snapshot --dev
```
Which automatically created a `snapshots.js` file in the root of the project.

## Opened cypress Launchpad and followed instructions to setup E2E testing
```
yarn run cypress open
```

The first time opening it, it should prompt something like `It looks like this is your first time using Cypress: xx.xx.xx`

And create a bunch of files:
- cypress.config.ts
- cypress/support/e2e.ts
- cypress/support/commands.ts
- cypress/fixtures/example.json

## Customized  `cypress.config.ts` 
- Disabled user agents, since they are dynamic and would influence our snapshots
- Disabled video recording. It’s not really useful anyway at this point
- Set the baseUrl of the `cy.visit` commands to the baseUrl of Docusaurus in dev

Here is the final config:
```
import { defineConfig } from "cypress";

export default defineConfig({
  userAgent: "mocked-user-agent",
  video: false,

  e2e: {
    baseUrl: "http://localhost:3000",
  },
});

```

## Customized  `package.json` 
Added a couple of scripts to make it easier to run cypress commands locally.
``` 
"scripts": {
  … 
  "cypress:open": "cypress open",
  "cypress:run": "cypress run",
  …
},
```

## Added a GitHub action to run e2e tests on push
Created a new action `.github/workflows/e2e.yml`
```
name: E2E Tests

on: [push]

jobs:
  cypress-run:
    runs-on: ubuntu-20.04
    steps:
      - name: Checkout
        uses: actions/checkout@v2

      - name: Install dependencies
        run: yarn install

      - name: Run Cypress tests
        uses: cypress-io/github-action@v4
        with:
          browser: chrome
          install: false
          start: yarn start
          wait-on: 'http://localhost:3000'
```

## Added some useful custom commands and tests
In `cypress/support/commands.ts` I've added:
- Loading of the snapshot plugin
- A custom command to set a fake cookie for Cookiebot
- A custom child Chainable command for DOM clickable elements. To prevent their default behavior.

### Objectiv commands
I've also edited `e2e.ts` in the same folder to load our own commands set, from the `objectiv_ommands.ts` file. We may move this in a plugin for Cypress eventually, bundling it with proper TS types and such, but for now I've left it here for easier testing.

Commands include:
 - `objectiv()`
 Initializes all other commands. Internally, it waits for the Trackers to be idle and returns an instance of RecordedEvents for further querying.
    
 - `clearEvents()` 
 Deletes all recorded events. May be useful when testing multiple concerns in the same session.
    
 - `filterEvents(options?: ((event) => boolean) | string[] | string)` 
 Filter events by their names, a list of their names or a predicate.
    
- `snapshotEvents(options?: { name?: string, json?: boolean })` 
Takes a snapshot of the list of recorded events and errors. Gets stored in `snapshots.js` in the root of the project. Relies on the `@cypress/snapshot` plugin.
    
- `startRecordingEvents()` 
Start recording events, if EventRecorder was stopped. May be useful when testing multiple concerns in the same session.
    
- `stopRecordingEvents()` 
Stop recording events, if EventRecorder was running. May be useful when testing multiple concerns in the same session.
        
- `withGlobalContext(name: string, id?: string)` 
Filter recorded events by a Global Context Name and, optionally, its identifier.
    
- `withLocationContext(name: string, id?: string)` 
Filter recorded events by a Location Context Name and, optionally, its identifier.

### Tests
And finally, in the `cypress/e2e` directory, Iv'e added a bunch of tests


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Run Cypress tests
+        uses: cypress-io/github-action@v4
+        with:
+          browser: chrome
```
sdirosa, line 18: In CI all browsers are run headless by default

cypress.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  userAgent: "mocked-user-agent",
+  video: false,
+
+  e2e: {
+    baseUrl: "http://localhost:3000",
+  },
+});
```
sdirosa, line 10: The userAgent changing over time is an issue for our `HttpContext`. In particular Cypress user agents contain the browser version and Cypress version. This means our snapshots would break any time a new version of either comes out.

I'm considering per-context serialization. This would allow us to serialize certain contexts in a different way when the serialization is invoked by the EventRecorder. 

It seems like this would also clean up a bit what currently we do in EventRecorder: `time` removal and `id` predictability

cypress/e2e/home_bottom_navigation.cy.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    // Visit home page
+    cy.visit('/');
+
+    // Scroll to footer
+    cy.get('.footer').scrollIntoView();
+
+    // Wait for star us notification to show up, it has an animation of 600ms
+    cy.wait(1000);
+    cy.get('div').contains('Star us on GitHub').should('be.visible');
```
sdirosa, line 20: I will look into some way of improving this.

Basically here we want to discard all events that occurred after the `cy.visit` and focus on what's gonna happen afterwards with the next actions.

Unfortunately there are async operations, such as the animation of the Star us notification in this case, that make this quite complex.

Things get even more complicated when testing the home page, since we have several requests to external services that may trigger in random order at random times in the first second.

I feel like it would be nice to have an API to wait for a certain event, or events, to have occurred. So we may easily clean up at certain points.

cypress/e2e/home_top_navigation.cy.ts
```diff
@@ -0,0 +1,34 @@
...
+    cy.get('a.navbar__brand').preventDefault().click();
+    cy.get('a').contains('About us').preventDefault().click();
+    cy.get('a').contains('Blog').preventDefault().click();
+    cy.get('a').contains('Jobs').preventDefault().click();
+    cy.get('a').contains('FAQ').preventDefault().click();
+    cy.get('a').contains('Docs').preventDefault().click();
+    cy.get('a.navGitHub').preventDefault().click();
+    cy.get('a.navSlack').preventDefault().click();
+    cy.get('a.navTwitter').preventDefault().click();
+    cy.get('a.navEmail').preventDefault().click();
```
sdirosa, line 28: Here I wanted to test all of these link from the home page, without having to navigate back and forth all the time. 
I created a custom command (`preventDefault`) to inject a "onclick" handler which prevents the default behavior.

This doesn't prevent our tracker event handlers to run, so we get a very fast test without side effects, which is a win-win.

cypress/fixtures/cookiebot.json
```diff
@@ -0,0 +1 @@
+{"widget":{"enabled":false,"position":{"position":"bottom-right","distanceVerticalPx":10,"distanceHorizontalPx":10},"theme":{"closed":{"backgroundColor":"#141414","foregroundColor":"#FFFFFF"},"opened":{"backgroundColor":"#FFFFFF","foregroundColor":"#333333","highlightColor":"#06A5FA","shadeColor":"#D6D6D6","changeBackgroundColor":"#5BC200","changeTextColor":"#FFFFFF","changeBorderColor":"#06A5FA","withdrawBackgroundColor":"#FFFFFF","withdrawTextColor":"#333333","withdrawBorderColor":"#06A5FA","detailsBackgroundColor":"#F2F2F2"}},"content":{"cultures":["en"],"default":"en","autodetectLanguage":true}}}
```
sdirosa, line 1: This is a random cookie content I grabbed from one of the runs. I don't think the content here matters. As long as the cookie exists I believe Cookiebot disables itself, even if the cookie is malformed.

To be tested a bit better perhaps.

cypress/support/e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
+
+Cypress.on('uncaught:exception', (err, runnable) => {
+  // we don't care about vimeo failing to load when that happens
+  if (err.message.includes('https://vimeo.com/670857141')) {
+    return false
+  }
+})
```
sdirosa, line 27: This tells Cypress to ignore 404 errors from Vimeo. Sometimes we purposely stub their API and return a 404 to speed up and remove side effects from tests running against the home page.

cypress/support/commands.ts
```diff
@@ -0,0 +1,79 @@
...
+Cypress.Commands.add('objectivClearEvents', () => {
+  cy.window().its('objectiv.EventRecorder').invoke('clear');
+  cy.objectivGetEvents().should('have.length', 0);
+})
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    objectivClearEvents(): Chainable<Subject>,
+    objectivGetEvents(): Chainable<Subject>,
+  }
+}
```
sdirosa, line 58: These commands may go in a plugin at some point, for now I'll leave them here for simplicity.

cypress/e2e/home_bottom_navigation.cy.ts
sdirosa: I've added a filtering options to the objectivGetEvents command, which allows us to easily filter out Events that are not relevant to a specific test/snapshot.

package.json
```diff
@@ -18,7 +20,7 @@
     "@docusaurus/core": "^2.0.0-beta.18",
     "@docusaurus/preset-classic": "^2.0.0-beta.18",
     "@mdx-js/react": "^1.6.22",
-    "@objectiv/tracker-react": "^0.0.20",
+    "@objectiv/tracker-react": "0.0.21-2",
```
ivarpruijn, line 23: let's update this before we merge it

sdirosa: So we wait until EOD?

sdirosa: Switched up to latest 0.0.21 and merged.

.github/workflows/e2e.yml
sdirosa, line 18: We can remove this if we upgrade cypress to 10.2

PR #468, opened 2022-06-22
Add validation video and collisions example to docs and blog post

ivarpruijn:
Adds the demo video about the new validation features to:
- The docs, under core-concepts > validation
- The blog post about the release, replacing the social image after the intro.

Also adds an example about collisions to the validation docs.

### DIFF
--- a/blog/2022-06-14 Release Helpful messages to debug your tracking instrumentation.md
+++ b/blog/2022-06-14 Release Helpful messages to debug your tracking instrumentation.md
@@ -13,6 +13,7 @@ authors: ivarpruijn
 
 import BlogImage from '@site/src/components/blog-image'
 import JoinSlackLink from '@site/src/components/join-slack-link';
+import VimeoPlayer from '@site/src/components/vimeo-player';
 
 [taxonomy]: https://objectiv.io/docs/taxonomy/
 [collector]: https://objectiv.io/docs/tracking/collector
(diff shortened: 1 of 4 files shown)


src/components/vimeo-player/index.js
```diff
@@ -14,6 +14,7 @@ import {
 function VimeoPlayer({ 
   videoId, 
   id, // required if the Video needs to be tracked
+  paddingBottom = "58.25%", // the ratio  for the video, e.g. 58.25% for 16:9 videos
```
ivarpruijn, line 17: Added to deal with different aspect ratios of videos, like the 2-minute intro vs these demo videos.

jansenbob (approved):
LGTM. I think we could take some steps in the recording of videos to make them a bit more suited for smaller screens though. I don't think you would be able to follow this with a small, sub par resolution screen.

sdirosa:
> LGTM. I think we could take some steps in the recording of videos to make them a bit more suited for smaller screens though. I don't think you would be able to follow this with a small, sub par resolution screen.

Another solution is to make the demo application used in the video public, so people can play around with it.

Or make a CodeSandbox / StackBlitz

ivarpruijn:
> 

That's 

> > LGTM. I think we could take some steps in the recording of videos to make them a bit more suited for smaller screens though. I don't think you would be able to follow this with a small, sub par resolution screen.
> 
> Another solution is to make the demo application used in the video public, so people can play around with it.
> 
> Or make a CodeSandbox / StackBlitz

Also a good idea, let's keep it in mind. Merging this for now.

src/components/vimeo-player/index.js
ivarpruijn: Good idea. Will make a note to see if that works later.

PR #478, opened 2022-06-28
Add validation and e2e testing video to docs

ivarpruijn:
- Adds a video showing the full flow from validation to end-to-end testing for frontend tracking instrumentation, right after the intro in the Validation section. 
- Keeps the specific video only for instrumentation as well.
- Adds an 'end-to-end testing' section linking to the bigger section on it, at the bottom.


vincenthoogsteder (approved):
LGTM

PR #501, opened 2022-07-21
E2E tests for docs

ivarpruijn:


.github/workflows/e2e-docs.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  cypress-run:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Go to docs directory
+        run: cd docs
```
sdirosa, line 13: Does this work?
I would try using the `working-directory` option instead, like we do in the analytics repo for tracker / backend.

I think what's may be happening is that the cypress command below is actually running against the website or something like that

ivarpruijn: That plus the fix for the dotenv config fixed it!

PR #518, opened 2022-08-11
Run tests only on push to corresponding paths

ivarpruijn:
Update workflows to only run docs tests for pushes to `/docs`, and website tests only for pushes _not_ to `/docs`.


sdirosa (approved):
Nice 🚢

PR #521, opened 2022-08-12
Swizzle top nav components to unbreak sticky behavior

ivarpruijn:


sdirosa (commented):
@ivarpruijn Tested this again with a fresh brain. The Sidebar seems to behave correctly now! 

Unfortunately the TOC still scrolls out of screen. I would expect the same behavior as on https://docusaurus.io/docs

I'm wondering why are these components not working out of the box? Shouldn't we report this issue? Or do you think it's related to our CSS customizations?

ivarpruijn:
> @ivarpruijn Tested this again with a fresh brain. The Sidebar seems to behave correctly now!
> 
> Unfortunately the TOC still scrolls out of screen. I would expect the same behavior as on https://docusaurus.io/docs
> 
> I'm wondering why are these components not working out of the box? Shouldn't we report this issue? Or do you think it's related to our CSS customizations?

I'll have a look at the TOC as well, thanks for catching that. The issue was due to our customization (adding another <nav> component before the menu, breaking the CSS), but I do think it might be an improvement for Docusaurus to not rely so strictly on the DOM structure here.

sdirosa:
> I'll have a look at the TOC as well, thanks for catching that.

Perfect, thanks 👌🏼 

> I do think it might be an improvement for Docusaurus to not rely so strictly on the DOM structure here.

Agreed. Seems a bit odd that customizing a few things will break such core functionality.

ivarpruijn:
> > I'll have a look at the TOC as well, thanks for catching that.
> 
> Perfect, thanks 👌🏼
> 
> > I do think it might be an improvement for Docusaurus to not rely so strictly on the DOM structure here.
> 
> Agreed. Seems a bit odd that customizing a few things will break such core functionality.

@sdirosa Fully swizzled the TOC (a tiny component anyway) and unbroke the sticky behavior.

sdirosa:
@ivarpruijn Totally works now. Very nice!

PR #535, opened 2022-08-29
Upgrade packages

ivarpruijn:
Upgrades Mermaid and several packages that have (not harmful) security warnings by Dependabot.


ivarpruijn:
Tested on staging, no issues found, so let's merge this also to silence the Dependabot warnings

PR #542, opened 2022-08-31
Attempt to limit code blocks furter for low-mem devices

ivarpruijn:


ivarpruijn:
Tested on staging, works, but doesn't seem to solve the issue on low-mem devices. Still doesn't hurt to merge.

PR #546, opened 2022-09-02
Experimental anonymous mode

ivarpruijn:
TODO:

- [x] Add e2e tests for anonymous mode.
- [x] Fix warning in console for docs.
- [ ] Upgrade to latest tracker.


docs/package.json
```diff
@@ -20,7 +20,7 @@
     "@docusaurus/preset-classic": "2.0.1",
     "@docusaurus/theme-search-algolia": "^2.0.1",
     "@mdx-js/react": "^1.6.21",
-    "@objectiv/tracker-react": "^0.0.23",
+    "@objectiv/tracker-react": "0.0.25-experimental.0",
```
sdirosa, line 23: You can upgrade 0.0.26-experimental.0: 

```yarn remove @objectiv/tracker-react && yarn add @objectiv/tracker-react@0.0.26-experimental.0```

File: src/theme/Root.tsx [hunk 2]
```diff
@@ -20,13 +21,17 @@ function Root({children}) {
...
     }))
   }
   
   // set tracker active or inactive based on cookie consent
   function setTrackerCookieConsent(cookieConsent) {
-    tracker.setActive(cookieConsent);
+    // set anonymous mode to reverse cookie consent
+    tracker.setAnonymous(!cookieConsent)
+    // always set tracker active
+    tracker.setActive(true);
```
sdirosa, line 34: This line should go, there's no need to touch the `active` option.

src/theme/Logo/index.tsx
```diff
@@ -24,8 +25,8 @@ export default function Logo(props: Props): JSX.Element {
   const {
     navbar: {title: navbarTitle, logo = {src: ''}},
   } = useThemeConfig();
- 
+
   const logoLink = useBaseUrl(logo.href || '/');
 
-  return <TrackedLink Component={OriginalLogo} id={'logo'} href={logoLink} title={navbarTitle} {...props} />
+  return <TrackedLinkContext objectiv={{Component: OriginalLogo}} id={'logo'} href={logoLink} title={navbarTitle} {...props} />
 };
```
sdirosa, line 32: I checked what's going on here and basically the OriginalLogo component uses some generic Props that are actually incompatible with the Element itself.

The Logo defines itself as an img tag, but then its Props specify that it has children. That makes no sense, as images are not container tags and cannot have children.

The second issue is that the Props type uses a legacy syntax for ref.

Since we don't use either, I just Omitted them from the original type. 

Here is a fixed (and cleaned up) version of this component:

```tsx
import React from 'react';
import type {Props} from '@theme/Logo';

import useBaseUrl from '@docusaurus/useBaseUrl';
import {useThemeConfig} from '@docusaurus/theme-common';

// OBJECTIV
import OriginalLogo from '@theme-original/Logo';
import { TrackedLinkContext } from '@objectiv/tracker-react';
// END OBJECTIV

export default function Logo(props: Omit<Props, 'children' | 'ref'>): JSX.Element {
  const { navbar: { title, logo } } = useThemeConfig();
  const logoLink = useBaseUrl(logo.href || '/');

  return <TrackedLinkContext 
    objectiv={{
      Component: OriginalLogo
    }} 
    id={'logo'} 
    href={logoLink} 
    title={title} 
    {...props}
  />
};


```

docs/package.json
```diff
@@ -20,7 +20,7 @@
     "@docusaurus/preset-classic": "2.0.1",
     "@docusaurus/theme-search-algolia": "^2.0.1",
     "@mdx-js/react": "^1.6.21",
-    "@objectiv/tracker-react": "^0.0.23",
+    "@objectiv/tracker-react": "0.0.26-experimental.0",
```
sdirosa, line 23: You can switch to version 0.0.26-experimental.0 for the latest and greatest version inclusive of all the feedback I got from Mike.

That version should also be appropriate to create and verify a snapshot for the test above.

src/theme/Logo/index.tsx
ivarpruijn: great, committed in this PR

docs/package.json
```diff
@@ -20,7 +20,7 @@
     "@docusaurus/preset-classic": "2.0.1",
     "@docusaurus/theme-search-algolia": "^2.0.1",
     "@mdx-js/react": "^1.6.21",
-    "@objectiv/tracker-react": "^0.0.23",
+    "@objectiv/tracker-react": "0.0.26-experimental.2",
```
sdirosa, line 23: I've re-published this version as: [0.0.26-next.0](https://www.npmjs.com/package/@objectiv/tracker-react/v/0.0.26-next.0)

sdirosa (approved):
lgtm!

PR #557, opened 2022-09-15
Add `display_sql_as_markdown` to example notebooks

ivarpruijn:
Adds `display_sql_as_md` operation to two example notebooks that didn't have it yet.


jansentom (approved):
LGTM, although it might be nicer to offer this through MH as this is the only import from Bach you'd need:
```
from modelhub import ModelHub, display_sql_as_markdown
modelhub = ModelHub(...)
df = modelhub.get_objectiv_dataframe(...)
display_sql_as_markdown(df)
```

ivarpruijn:
> LGTM, although it might be nicer to offer this through MH as this is the only import from Bach you'd need:
> 
> ```
> from modelhub import ModelHub, display_sql_as_markdown
> modelhub = ModelHub(...)
> df = modelhub.get_objectiv_dataframe(...)
> display_sql_as_markdown(df)
> ```

Much nicer, didn't know that works too. Think we should do the same in all (regular) notebooks where we only use modelhub, will pick that up with the rest of the changes I'll be working on.

PR #576, opened 2022-10-05
Move core concepts from tracking docs to tracking root and other places

ivarpruijn:


sdirosa (approved):
Much nicer structure than before 👍🏼

PR #580, opened 2022-10-06
Add steps to Get Objectiv Up page

ivarpruijn:


docs/docs/home/up.md
```diff
@@ -61,10 +59,9 @@ Initial password:
 metabase1
 ```
 
-After the first login, you can change these details [here](http://localhost:3000/account/profile).
-
 #### Creating BI Dashboards
```
ivarpruijn, line 62: This section should move to another page under 'Modeling', but will tackle that in separate PR, as we'll need to update both the MH/Bach docs and Jupyter notebooks (plus add redirects). After that, this section can just be a link to that page.

PR #589, opened 2022-10-24
change path to generated iglu schema's

borft:
iglu schema's are now generated from the /schema dir, this updates the link in the docs


ivarpruijn (approved):
LGTM

